PR #22, opened 2014-12-01
Modecfg interoperability fixes

lkundrak:


lkundrak:
Moved here https://github.com/libreswan/libreswan/pull/40

PR #23, opened 2015-01-05
If the build fails (really) abort.

cagney:
if ! make ; then ... looses the exit status leading to the script
always succeeding.


letoams:
this has been merged in? :)

PR #25, opened 2015-03-20
Simplify Upstart job, use exec in script section so PID is tracked correctly

CameronNemo:
The sections are always in root, no need to cd to root. $PLUTO_OPTIONS are not used in the pre-start command so there is no need to source the default file there. The script section should use exec so that the correct PID is tracked (this is not too much of an issue, since the PID tracked formerly was the shell, which exits when the real daemon does).


letoams:
merged, will be in 3.13.

Thanks!

PR #30, opened 2015-05-11
Update ipsec.conf.in

aabdnn:
Without the "exec", you have a shell hanging around, and upstart thinks the shell's PID is the main process. This may confuse the pre-stop part. If you "exec" pluto, then it takes over the PID of the shell that starts it, and upstart monitors the correct PID. Then, when it is shut down in the pre-stop, upstart should not try to restart it. Let me know if it works.

### DIFF
--- a/initsystems/upstart/ipsec.conf.in
+++ b/initsystems/upstart/ipsec.conf.in
@@ -22,7 +22,7 @@ script
     @FINALSBINDIR@/ipsec --checknss
     @FINALSBINDIR@/ipsec --checknflog
     @FINALSBINDIR@/ipsec _stackmanager start
-    @FINALLIBEXECDIR@/pluto --config @FINALCONFFILE@ --nofork $PLUTO_OPTIONS
+    exec @FINALLIBEXECDIR@/pluto --config @FINALCONFFILE@ --nofork $PLUTO_OPTIONS
 end script
 
 pre-stop script


arhbkb:
I don't understand it... the same sequence works with BIND, for example. If I call "rndc stop" from within a pre-stop script, upstart runs "rndc stop", which causes BIND to exit, and then upstart goes into the "stop/waiting" state, as I would expect.

Just to be sure, could you "start ipsec", and then "status ipsec" and compare the PID reported by upstart with the PID pluto is running with? I don't know enough about libreswan to know if other things are going on behind the scenes.

letoams:
that all sync's up

 stop ipsec
ipsec stop/waiting
start ipsec
ipsec start/running, process 2241
 status ipsec
ipsec start/running, process 2241
 pidof pluto
2241

arhbkb:
Well, at the moment, I don't know what else to suggest. I'm in the middle of RIPE 70, with lots to do, so I'm afraid I don't have much time to debug this much further. If you still haven't found a solution next week, I may be able to have another look.

letoams:
On Tue, 12 May 2015, Anand Buddhdev wrote:

> Well, at the moment, I don't know what else to suggest. I'm in the middle of RIPE 70, with lots to do, so I'm afraid I don't
> have much time to debug this much further. If you still haven't found a solution next week, I may be able to have another
> look.

Enjoy RIPE. The workaround for now is to let upstart send signals to the
main daemon.

letoams:
Is this still a problem for you in libreswan-3.16 ?

aabdnn:
This was never a problem for me. You had asked for suggestions to manage pluto on CentOS 6, so I was providing suggestions. Did you ever find a solution?

letoams:
We have our _plutorun wrapper for sysvinit systems that handles restarting (or not restarting)

PR #33, opened 2015-07-13
twofish: Fix CALC_SB_* macros

lkundrak:
They were used as loop bodies, causing only the first line to be iterated.


letoams:
Thanks for the patch. I'll do the interop testing. I had done some testing on the existing code and it clearly did not work as it should.

letoams:
twofish interop confirmed again - this has fixed the twofish problem. Merged in for 3.14 release. Thanks!

PR #38, opened 2015-10-28
_updown.*: Fix NetworkManager callback

lkundrak:
Letting NM know is not just specific to resolver configuration management, NM
always needs to know if setting up the connection succeeded or timed out.

Moreover, it's probably not a good idea to unset the variables upon disconnect
as it would be nice if NetworkManager could identify the connection that goes
down.


letoams:
merged - thanks. although we should fix the ipv6 handling too

PR #39, opened 2015-10-28
systemd: add socket activation

lkundrak:
This fixes a startup race where the tools don't know whether it's safe to use
the managemenet socket after launching the service.


letoams:
merged, thanks

PR #40, opened 2015-10-28
Modecfg interoperability fixes

lkundrak:


letoams:
pulled in and will be in 3.17. Thanks!

PR #45, opened 2015-11-25
KLIPS: kernel 4.1 requires HAVE_USER_NS

roelvanmeer:
Newer linux kernels always use the same interface to uids, whether they use
user namespaces or not.

This solves a build failure on linux 4.1.13 where the config dit not include CONFIG_USER_NS.
I haven't checked kernel versions before 4.1.


letoams:
Merged in, will be in 3.16. Thanks

PR #51, opened 2016-03-17
Fix build with new GCC

lkundrak:


letoams:
On Thu, 17 Mar 2016, Lubomir Rintel wrote:

>   https://github.com/libreswan/libreswan/pull/51
>
>     Commit Summary
>
>  *  constants: fix build with gcc6
>  *  pluto: fix a typo

Thanks. I'm a little worried some of those structs are now unused, so
I'm going to have a closer look at what's going on before removing them.

Paul

letoams:
fixed slightly differently, but thanks for the pointer.

PR #52, opened 2016-03-29
Log phase 2 cipher mismatches.

solemnwarning:
Log phase 2 cipher mismatches rather than hiding them away in the debug messages.

Mismatches in phase 1 already get logged in this way.


letoams:
merged in. thanks!

PR #53, opened 2016-04-19
Fix addconn verbose output

doctaweeks:
This fixes an inverse logic bug when addconn is run with verbose output.
Introduced by 4fc07a76d91d502918abbd13c7c213905f7a4e41


letoams:
Thanks. pulled in

PR #55, opened 2016-04-19
Fix for #54

### LINKED ISSUE #54: resolve_defaultroute_one fails when a conn uses %defaultroute and a named peer
When resolve_defaultroute_one is called for a connection that has left=%defaultroute and right=<dns name>, it determines that it does not have an IP for the peer, source, or gateway. However, under this set of condition it can not populate any of these field and fails to bring up the resulting connection ("We cannot identify ourselves [...]"). If either left or right is replaced with an appropriate IP address the connection will succeed.

doctaweeks:
This should fix #54 where neither left nor right has a configured IP address. Without at least one a route can not be determined.

### DIFF
--- a/programs/addconn/addconn.c
+++ b/programs/addconn/addconn.c
@@ -443,7 +443,7 @@ static int resolve_defaultroute_one(struct starter_end *host,
 				r_gateway,
 				r_interface,
 				r_source, rtmsg->rtm_table,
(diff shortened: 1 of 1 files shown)


doctaweeks:
Sorry - forgot to rebase back to master but it looks like the other patch has been included anyway.

letoams:
Merged. Thanks!

PR #58, opened 2016-05-25
Follow connaddrfamily when resolving addresses

doctaweeks:
This fixes a bug where the v6 address for a hostname was incorrectly used when connaddrfamily=ipv4.


letoams:
Merged, and will be in 3.18. Thanks!

PR #62, opened 2016-06-20
section intro line needs whitespace between "type" and "name"

dkg:
xmlto 0.0.25-2 (debian stable) apparently elides the space between
"type" and "name", which makes this hard to read in the output text.
(i.e. it looks like "typename")

Also, we want this directive to always be on a single line no matter
how crazy narrow the terminal is.

This non-breaking space should solve both problems.


letoams:
merged, will be in 3.18

PR #64, opened 2016-06-20
avoid bashism and other risks of echo

dkg:
-e and -n are arguments to /bin/echo and work in bash's builtin echo.

Other shells, like posh and dash, do not support -e and -n.

With /bin/sh supplied by dash, i see:

```
0 dkg@alice:~$ /usr/sbin/ipsec --help
Usage: ipsec <command> <argument ...>
where <command> is one of:

	start-e -n
stop
	restart-e -n
status
	import-e -n
initnss
	checknss-e -n
checknflog
	addconn-e -n
auto
	barf-e -n
ikeping
	look-e -n
newhostkey
	pluto-e -n
readwriteconf
	rsasigkey-e -n
setup
	showhostkey-e -n
verify
	whack

See also: man ipsec <command> or ipsec <command> --help
See <https://libreswan.org/> for more general info.
Linux Libreswan 3.18dr2 (netkey) on 4.5.0-2-amd64
0 dkg@alice:~$
```

Also, when supplying variables as the first part text to echo, there
is a risk that the parameters may start with a -.  In that case, some
implementations of echo will treat the entire string as arguments,
which will likely fail.

Using printf instead resolves these problems.  This patch fixes ipsec
and newhostkey.  It would probably be wortwhile to scan through the
rest of the code to fix issues like this (perhaps just uniformly
remove echo in most places?), but i have not had time to do so.

### DIFF
--- a/programs/ipsec/ipsec.in
+++ b/programs/ipsec/ipsec.in
@@ -124,21 +124,21 @@ case "${1}" in
 		GOTTWO="${f}"
 	    else
 		# second of two entries, we can print
-		echo -n "	${GOTTWO}"
+		printf "\t%s" "${GOTTWO}"
 		if [ "${#GOTTWO}" -ge 16 ]; then
-		    echo -e -n  "\t"
+		    printf  "\t"
 		elif [ "${#GOTTWO}" -ge 8 ]; then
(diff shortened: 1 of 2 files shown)


letoams:
has been merged in

PR #65, opened 2016-06-23
Nssdir cleanup

dkg:


letoams:
This patch misses several items where nss dir is used. I'll see if I can pick this up later today

dkg:
On Thu 2016-06-23 09:34:59 -0400, Paul Wouters (libreswan) wrote:
> This patch misses several items where nss dir is used. I'll see if I can pick this up later today

Any word on this?  Even if it's not complete, it's much closer to
complete than the status quo documentation :)

         --dkg

letoams:
Merged in, will be in 3.19

PR #70, opened 2016-08-08
Enable "systemctl help ipsec"

dkg:
The Documentation= key in a [Unit] section of a systemd .service file
is used by systemctl to provide an easy way for an administrator to
get access to documentation about a given service.  Search for
Documentation= in systemd.unit(5) for more details.


letoams:
merged

PR #74, opened 2016-09-08
Add 'libsystemd-dev' in Debian pkg list.

theznx:
Hello,
On Debian Jessie build fail,
<systemd/sd-daemon.h> is need and found with libsystemd-dev package.


letoams:
merged, thanks

PR #78, opened 2016-11-30
Include strerror in the error message when file open fails.

rthille:
I was seeing a weird error when running with multiple config
files. Never did track it down because my rebuilt part of libreswan
wouldn't run with the rest of the distro's version. Was easier to just
switch to using a single config file.


letoams:
thanks, merged and will be in 3.19

Issue #80, opened 2017-01-12
The vti interface vanishes when peer goes down and up


yu-shiba:
I was build and tested route based vpn(libreswan <-> libreswan) from master branch source. 
```
    left=x.x.x.x
    leftid=y.y.y.y
    leftsubnet=0.0.0.0/0
    right=z.z.z.z
    rightid=z.z.z.z
    rightsubnet=0.0.0.0/0
    mark=1/0xffffffff
    vti-interface=vti01
    vti-routing=no
    vti-shared=no
    leftvti=169.254.252.XXX/30
```
Normal startup is not a problem.
But, when the connection peer goes down and up, the vti interface leave deleted.
This situation is fixed unless it is restarted.

In a little examination, _updown.netkey script receives the **down-client**, script got PLUTO_CONN_KIND was 'CK_PERMANENT'.
I wonder if we do not consider this situation.

letoams:
if the connection goes down because the remote peer stops responding, it is expected that we tear down the tunnel, and the VTI interface.

But once the remote peer comes back again, the tunnel is expected to re-establish and the VTI interface should be re-added again.

So I dont think your patch is correct. When the connection is not shared and the connection is a permanent, then there should not be other users of the VTI device, and so deletion seems right?

But perhaps because we are going down and then up perhaps we are not properly re-adding the VTI device

yu-shiba:
I thought so. But it remained deleted.

I tested it was route based vpn(*subnet=subnet=0.0.0.0/0), it related to it?
The addition of vti is executed at **prepare-host** or **prepare-client**, but will not it be triggered at route based vpn.

letoams:
On Fri, 24 Mar 2017, Yuichiro SHIBAYAMA wrote:

> I tested it was route based vpn(*subnet=subnet=0.0.0.0/0), it related to it?
> The addition of vti is executed at prepare-host or prepare-client, but will not it be triggered at route based
> vpn.

If you have a remote subnet of 0.0.0.0/0 then you are basically setting
up a routing based VPN, so that means you are expected to manually
route anything you want to pass through the VPN (and you cannot set
vti-routing=yes because you cannot route 0.0.0.0/0 over the VTI
device without imploding your connectivity.

Paul

yu-shiba:
Since I want to route dynamically using bgp, and chosed "route based vpn", but are you using the wrong way?

I finally want to make a connection with the following.
- AWS VPC
- Yamaha vpn router (RTX 1200)
[...]

PR #81, opened 2017-01-12
Patch for #80

yu-shiba:
- Exclude CK_PERMANENT on delvti.
- Change up-client to maintain symmetry with down-client


letoams:
thanks. It did not make it for 3.19, but we are looking at adding a proper testcase and getting a fix in.

letoams:
We have left out the deletion of VTI interfaces completely and will re-implement this with proper reference counting

PR #89, opened 2017-01-26
close parens

dkg:
i'm pretty sure i suggested this fix with a proper close-paren initially :)


letoams:
oops. sorry :)
fixed.

PR #110, opened 2017-07-17
Fix build with gcc-7.1

lkundrak:
Did a master build on fresh Fedora 26, failed with a couple of warnings triggering -Werror.


letoams:
merged in, thanks

PR #119, opened 2017-08-11
fix speling

dkg:


letoams:
Has been pulled in by Antony

PR #122, opened 2017-08-24
Specify length of F4 exponent to 3

tisj:


letoams:
Solved slightly differently in master, bug fix is in git master and will be in 3.22

PR #131, opened 2017-10-23
Prefer use of echo to printf, so escape sequences are interpreted correctly.

### LINKED ISSUE #130: printf is incorrect in resolv.conf generation
2a2578b5763d027c0f5723f4c66844a801f710d4 added "%s" as the first argument to `printf` when generating resolv.conf files, however both coreutils printf and bash built-in printf ignore escape sequences in string arguments. This breaks the generated resolv.conf files, as the newlines are printed as "\n". Instead of using printf at all, `echo -e` would probably be sufficient.

afranchuk:
This closes #130.


letoams:
fixed slightly differently, thanks!

PR #150, opened 2017-12-19
Small bug fix in ikev2_parent.c

vukasink:
notifies variable was in one if branch decremented after the evaluation of np (next payload) variable, and
it should be decremented before it.


libreswan:
merged, thanks!

PR #153, opened 2018-01-18
logging: dump NO_PPK_AUTH octets in PRIVATE

vukasink:
Changed logging from DBG_CONTROL to DBG_PRIVATE, following the same
principle used for "normal" PSK octets


libreswan:
merged, thanks!

PR #157, opened 2018-01-26
logging: fixed unclear log

vukasink:
At this point there isn't any loaded PPK, and so no PPK_ID. Maybe this log was meant for the responder side?


letoams:
merged, thanks!

PR #160, opened 2018-02-02
Update virtual.c

systemcrash:
Attempt to be slightly more helpful at this error condition.


letoams:
ivate network virtual IP was require

PR #161, opened 2018-02-02
Update virtual.c

systemcrash:
exclusion form for IPv6


letoams:
thanks merged all the virtual.c changes into one

PR #162, opened 2018-02-02
Update virtual.c

systemcrash:
make error strings more human readable ;)


letoams:
thanks merged all the virtual.c changes into one

PR #178, opened 2018-05-15
remove include in6.h

ouhaiyun0:


ouhaiyun0:
@ouhaiyun0
remove include in6.h

PR #184, opened 2018-06-01
Fix a segfault of addconn caused by free() of invalid memory

brozs:
The addconn program crashed on connection configuration syntax errors.
E.g. on duplicit "left" or "leftid" keywords.


letoams:
merged, thanks!

Issue #170, opened 2018-04-02
Bug - Libreswan 3.2x/master with Linux kernel 4.14.x


Pierre-A3i:
![kp-4 14 34](https://user-images.githubusercontent.com/34443850/38893737-7db3c196-428b-11e8-9f40-5915296c905e.jpg)

Hi, 
Libreswan 3.2x/master are not working with Linux kernel 4.14.x and KLIPS.
Tunnel is build fine but there is no communication. The cryptographic lookup seems failed.
With  "ipsec stop / restart" I get kernel panic above.

KLIPS Debug Log ( Kernel 4.14.x):  

May  5 11:52:57 Blackbox-CRE kernel: [  583.449883] klips_info:ipsec_init: KLIPS startup, Libreswan KLIPS IPsec stack version: 3.20
May  5 11:52:57 Blackbox-CRE kernel: [  583.450081] NET: Registered protocol family 15
May  5 11:52:57 Blackbox-CRE kernel: [  583.450630] klips_info:ipsec_alg_init: KLIPS alg v=0.8.1-0 (EALG_MAX=255, AALG_MAX=255)
May  5 11:52:57 Blackbox-CRE kernel: [  583.450632] klips_info:ipsec_alg_init: calling ipsec_alg_static_init()
May  5 11:52:57 Blackbox-CRE kernel: [  583.456616] KLIPS: lookup for ciphername=cbc(aes): not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.461761] KLIPS: lookup for ciphername=cbc(twofish): not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.466707] KLIPS: lookup for ciphername=cbc(serpent): not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.469024] KLIPS: lookup for ciphername=cbc(cast5): not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.471425] KLIPS: lookup for ciphername=cbc(des3_ede): not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.473854] KLIPS: lookup for ciphername=cipher_null: not found
May  5 11:52:57 Blackbox-CRE kernel: [  583.476595] KLIPS cryptoapi interface: alg_type=14 alg_id=2 name=hmac(md5) ctx_size=40 keyminbits=128 keymaxbits=128, found(0)
[...]

PR #190, opened 2018-06-25
KLIPS: Fix HMAC destroy - HMAC calculation using ahash

tisj:
Fixes issue with destroy of HMAC keys, see #170 https://github.com/libreswan/libreswan/issues/170

Additionally also fixes HMAC calculation (when using ahash), by correcting the length of the input buffer.


letoams:
merged, sorry for missing parts in #189

PR #191, opened 2018-06-27
Clean up README.nss

dkg:
We're using the sqlite variant of the NSS db, README.nss should refer
to it consistently.


letoams:
merged, thanks

PR #193, opened 2018-06-29
Added spacing in log output.

ttimpe:
Made output of error more uniform. This was most likely just an oversight.


libreswan:
merged

PR #194, opened 2018-06-29
Remove duplicated "--rereadsecrets" option from ipsec_auto man page

fgiudici:
Found the "--rereadsecrets" option appeared twice in the ipsec_auto man page... just removed the duplicated one.


libreswan:
merged, will be in 3.26

PR #203, opened 2018-09-20
Topic/netlink eroute connection delete with SA marks

tisj:
When a tunnel is removed, the following calls to netlink eroute raw emerge:

1. netkey eroute sadb_op 0x2 (replace), esatype 0x3d 'eroute_connection replace with shunt'
2. netkey eroute sadb_op 0x6 (del inbound), esatype 0x0 'unk255.10000@<IP address>'
3. netkey eroute sadb_op 0x3 (del), esatype 0x3d 'eroute_connection delete'
4. netkey eroute sadb_op 0x6 (del inbound), esatype 0x3d 'eroute_connection delete inbound'

The third step would produce an error, such that xfrm policy (outbound) is not actually deleted, an error is produced and the fourth step wouldn't take place. 

To my believe due to missing the (necessary) SA marks, as I was able to remove the manually using ip xfrm policy. And it turned out right.

I am not sure what the impact is of allowing IPSEC_POLICY_DISCARD: can it occur in other circumstances when sadb_op is not ERO_DELETE


letoams:
Confirmed (using test case netkey-vti-01) and merged, with some indenting changes. will be in 3.27. Thanks!

PR #205, opened 2018-09-28
Cleanup

dkg:
this is minor cleanup: a few orthographic fixes, and changes of many links that were http:// to use https://

there are still more http:// links remaining in the codebase, but significantly fewer than before. it'd be great if someone wanted to try to reduce the number further.


letoams:
merged, will be in 3.27. thanks!

letoams:
Just a note. We realised the commit actually moved the license links to v3+ while our license is v2+, so another commit addresses this.

dkg:
the license links already pointed to v3+, if you followed all the HTTP 302 redirections -- glad this spurred you make the corrections :)

letoams:
But we are v2+ not v3+

PR #210, opened 2018-10-09
Fix ocsp-method=post

brozs:
This patch attempts to fix the "ocsp-method=get|post" selection that had no effect after switching to NSS CERT_PKIXVerifyCert().

Previously, the OCSP GET method was used even with "ocsp-method=post" set in ipsec.conf. With this patch, only the OCSP method POST is used for validation when configured so.

It is done so by setting the CERT_REV_M_FORCE_POST_METHOD_FOR_OCSP flag in NSS.

I would appreciate a code review, any feedback welcomed.

### DIFF
--- a/programs/pluto/fetch.h
+++ b/programs/pluto/fetch.h
@@ -25,3 +25,4 @@ extern long curl_timeout;
 extern bool crl_strict;
 extern bool ocsp_strict;
 extern bool ocsp_enable;
+extern bool ocsp_post;
--- a/programs/pluto/nss_cert_verify.c
+++ b/programs/pluto/nss_cert_verify.c
@@ -196,7 +196,7 @@ static void set_rev_per_meth(CERTRevocationFlags *rev, PRUint64 *lflags,
 	rev->chainTests.cert_rev_flags_per_method = cflags;
 }
 
-static unsigned int rev_val_flags(PRBool strict)
+static unsigned int rev_val_flags(PRBool strict, PRBool post)
 {
 	unsigned int flags = CERT_REV_M_TEST_USING_THIS_METHOD;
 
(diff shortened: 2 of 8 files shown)


letoams:
I checked this on nic using:

 tcpdump -s 0 -A 'tcp dst port 2560 and (tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x504f5354)'

But now it always shows POST even if the ocsp-method=get is set

letoams:
wore, it shows:

POST / HTTP/1.0
Host: nic.testing.libreswan.org:2560
Content-Type: application/ocsp-request
Content-Length: 100

I think it should be HTTP 1.1 to support POST ?
Also Host: should not have a port number ?

brozs:
Thank you for your time to review. All this pull-request does is setting a flag for the NSS library, everything else is done within NSS internally.

With `ocsp-method=get` the first attempt will be made using the GET method, if that fails without a "HTTP 200 OK" response(e.g. after the timeout expires), it will retry using POST. This retry cannot be used with the built-in OCSP responder in openssl, it stops with "Invalid request" error on receiving a GET request.

With `ocsp-method=post` only the POST method will be used. I have verified it's working using the built-in OCSP responder in openssl:

```
openssl ocsp -port 80 -text -sha256 -index intermediate/index.txt -CA intermediate/certs/ca-chain.cert.pem -rkey <private_key.pem> -rsigner <ocsp_signer_cert.pem>
```

Additionally, the ipsec.conf manual page says:
```
       ocsp-method
           The HTTP methods used for fetching OCSP data. Valid options are get
           (the default) and post. Note that this behaviour depends on the NSS
           crypto library that is actually performing the fetching. When set
           to the get method, post is attempted only as fallback in case of
           failure. When set to post, only the post method is ever used.
```
The ocsp-method=post has no effect on the behaviour without setting the appropriate flag in NSS (`CERT_REV_M_FORCE_POST_METHOD_FOR_OCSP`) as using `CERT_ForcePostMethodForOCSP(TRUE)` does not apply when `CERT_PKIXVerifyCert()` is used for certificate verification.

A next step might be simplifying the libreswan's code and removing the parts where forcePost is being set.

Regarding the HTTP protocol itself, it's using the default simple HTTP client internal to NSS (`pkix_pl_httpdefaultclient.c`), I made no changes in that matter.

But anyway, for completeness:

The HTTP POST method is valid in HTTP 1.0 as per RFC 1945.

The Host header as per RFC 2616 is defined as:
```
Host = "Host" ":" host [ ":" port ]
```
And more specifically:  _A "host" without any trailing port information implies the default port for the service requested (e.g., "80" for an HTTP URL)._

The internal HTTP client of the NSS library omits the port in the Host header when it's equal to "80".

letoams:
On Tue, 16 Oct 2018, Štěpán Brož wrote:

> Thank you for your time to review. All this pull-request does is setting a flag for the NSS library, everything else is done within NSS
> internally.
> 
> With ocsp-method=get the first attempt will be made using the GET method, if that fails without a "HTTP 200 OK" response(e.g. after the
> timeout expires), it will retry using POST. This retry cannot be used with the built-in OCSP responder in openssl, it stops with "Invalid
> request" error on receiving a GET request.

I will retest to ensure I see a first GET request.

Paul

brozs:
I would like to know if you plan to actually merge this at some point? This patch is fixing a bug of libreswan.

letoams:
On Wed, 21 Nov 2018, Štěpán Brož wrote:

> I would like to know if you plan to actually merge this at some point? This patch is fixing a bug of
> libreswan.

Oops. Thanks for reminding me. I have to rerun the tests to confirm your
comments, and then it will be merged in. Sorry for the delay.

Paul

letoams:
merged. thanks!

PR #212, opened 2018-10-15
Automatically detect whether Curve25519 is available in NSS for USE_DH31

dkg:
currently, USE_DH31 defaults to true.

On some platforms, where older versions of NSS are in use, this should
default to false.

This patch automatically detects such systems and disables USE_DH31 on
those platforms.

It produces some amount of noise on stderr about CURVE_25519 during
the build on those older platforms, but that seems ok to me, since
those platforms should know that they're missing a modern feature.

If you prefer less noise, i'd be happy with any modification that does
a similar sort of autodetection.

### DIFF
--- a/mk/userland-cflags.mk
+++ b/mk/userland-cflags.mk
@@ -240,7 +240,7 @@ ifeq ($(USE_DH24),true)
 USERLAND_CFLAGS += -DUSE_DH24
 endif
 
-USE_DH31 ?= true
+USE_DH31 ?= $(shell if printf '\#include <secoidt.h>\nint main() { return SEC_OID_CURVE25519; }\n'| $(CC) -x c $$(pkg-config --cflags nss) -o /dev/null -; then echo true; else echo false; fi )
 ifeq ($(USE_DH31),true)
 USERLAND_CFLAGS += -DUSE_DH31
 endif


letoams:
this would break cross compiling. I'd rather see a method that depends on defines in nss headers

mcr:
You can't use defines in the headers, because the libnss that is on the
running system might not be the same.  That's the whole point of pluggable
systems.  So if the system compiling has defines that let you identify
Curve25519, then you can compile support in, but then you have to check at
runtime to see if the system has that support and turn it off it it's not
there.

dkg:
@mcr if you use the defines in the headers, then the built package will explicitly depend on a version of the NSS library that does export the symbol used.

if your package tracking system can't ensure that those dependencies are properly met (e.g. by ensuring that a recent-enough version of NSS is available to install such a built version of libreswan), then the fault lies in the package tracking system, not the compilation process.

dkg:
@letoams the patch i proposed does indeed depend on defines in the NSS headers, based on its use of `pkg-config --cflags nss`.   Isn't pkg-config supposed to support cross compiling appropriately, if the appropriate cross-platform nss development files are installed?

antonyantony:
> @letoams the patch i proposed does indeed depend on defines in the NSS headers, based on its use of `pkg-config --cflags nss`. Isn't pkg-config supposed to support cross compiling appropriately, if the appropriate cross-platform nss development files are installed?

+1  

I tested this on F28 cross compiling OpenWRT for AArch64 . 
The test  seems to work. ie. I see  -DUSE_DH31 on compile line. And run time show group=DH31

First I  removed nss on the host.  " dnf remove nss-util-devel" 
then compiled. 

However, this is an ugly long line in the userland-cflags.mk:)  I wonder if there some way make this line pretty looking:)

letoams:
On Thu, 6 Jun 2019, Antony Antony wrote:

>       @letoams the patch i proposed does indeed depend on defines in the NSS headers, based on its use of
>       pkg-config --cflags nss. Isn't pkg-config supposed to support cross compiling appropriately, if the
>       appropriate cross-platform nss development files are installed?
> 
> +1
> 
> I tested this on F28 cross compiling OpenWRT for AArch64 .
> The test seems to work. ie. I see -DUSE_DH31 on compile line. And run time show group=DH31
> 
> First I removed nss on the host. " dnf remove nss-util-devel"
> then compiled.

Did you have to remove nss on the host to avoid confusing with
pkg-config ?

antonyantony:
> Did you have to remove nss on the host to avoid confusing with pkg-config ?

no. I tested with and without nss/nspr rpms. I imagined without NSS rpms are more important test when corss compiling.
 
FYI: this patch is part of Debian  and OpenWRT.

Issue #211, opened 2018-10-15
http://www.freeswan.org (hosted by libreswan) does not offer https:// 


dkg:
It looks like http://www.freeswan.org/ is currently hosted by the libreswan project.

however, going to https://www.freeswan.org/ gives a certificate error, because an https://www.libreswan.org certificate is used.

Please enable https for https://www.freeswan.org/ as well!

bleve:
Thank you, fixed.

dkg:
fix confirmed.  thanks for the prompt action!

PR #213, opened 2018-10-15
Https cleanup

dkg:
more http→https cleanup.  note that #211 ought to be fixed before merging 0186ee3


dkg:
@bleve fixed #211 -- so i think these changes are now safe to merge.

letoams:
I picked most of the ones listed in your patch set, but not all of them:

- Anything in CHANGES files that is not the libreswan CHANGES, as those are historic files
- Skipped anything in linux/ since that entire directory is about to be removed with KLIPS removal
- security/ as those are copies of the CVE releases done and I don't want to modify those

PR #217, opened 2018-10-17
sort all wildcarded object files for reproducibility

dkg:
Without this sort, the linker may assemble the final object files in
the order in which their source files are found in the filesystem (or
however make chooses to interpret the $(wildcard ...) function).

By sorting the results lexicographically, we aim to assemble
byte-identical output files, regardless of the filesystem order or
other possible non-determinism introduced by make.

(i've applied this patch to the debian packaging of libreswan between versions 3.27-3 and 3.27-4;  results: 3.27-3 is non-deterministically unreproducible, and 3.27-4 appears to be reliably reproducible)


letoams:
merged, thanks!

PR #235, opened 2019-01-30
If cert_VerifySubjectAltName fails, be sure to log what it was (the name) that the code was looking for

kmansoft:
Rationale: from watching the mailing list, it seems somewhat difficult to get just the right stuff into your
certificates' subjectAltName fields (or at all) for certificate validation to pass.

Right now when cert_VerifySubjectAltName fails, it just logs `No matching subjectAltName found` giving no indication of what actual name it was trying to match.

More logging should make it easier for users to troubleshoot such issues:

`No matching subjectAltName found, was looking for <actual name or address here>`


letoams:
Thanks, I committed a patch similar to this that also fixes a DBG -> log message

PR #239, opened 2019-03-23
whack: fix non operational connection flags / arguments

fcntl:
- Some flags for whack connection setup are not working properly. This is caused by missing option offset.
- The field "nat_keepalive" is defined as enabled by default but initialized disabled.
- Help description of program arguments is incomplete.


letoams:
merged. Thanks!

PR #247, opened 2019-06-02
minor: Fix spelling and grammar

dkg:
therefor → therefore
transfered → transferred
acceptible → acceptable

This also involves some clean up of "allow to", the use of which
typically presents some amount of confusion about who is doing the
allowing, and who is being allowed.  Given that the options in
question are all called accept-*, it makes more sense to describe them
as "accepting" anyway.

Signed-off-by: Daniel Kahn Gillmor <dkg@fifthhorseman.net>


letoams:
merged. thanks!

PR #281, opened 2019-10-22
Modified verify.in to make it python3 compatible 

AnandBibhuti:
Modified verify.in to make it python3 compatible


letoams:
Thanks. Applied your fix with a minor modification.

PR #290, opened 2019-11-09
Xauth cleanup

avis:
add pam session?!


avis:
Approved by mgibson

letoams:
Can you explain what the patch is trying to do?

It seems you are trying to hold open the pam session per connection, but
for roadwarriors, the connection is instantiated into a new connection,
so I'm not sure your approach will do much ?

Paul

PR #310, opened 2020-02-12
Documentation:correct spelling mistake

andmeics:
Modified show.in to correct spelling mistake


bleve:
This patch looks completely wrong - why are you removing most of the command?

bleve:
Thank you for reporting issue. Fixed with different change.

PR #311, opened 2020-02-19
account for fragmentation on clients with small MTU sizes

johnmah:
Some Windows clients will fragment IKE payloads on a ridiculously low MTU sizes (we have observed as low as 626 bytes in the field, although I would not be surprised to see 576). This can cause pluto to drop large IKE_AUTH packets with fragment counts of >16.

This often gets reported to Windows clients as error code 809 as the server never responds to the IKE_AUTH, despite receiving all of the fragments.


letoams:
was merged

PR #314, opened 2020-03-17
Fixing manpage of pluto.

tanuja3:
Fixed the single inverted commas.
 They were displaying as "doesn*(Aqt" instead of "doesn't".

Signed-off-by: tanuja3 <dtanujakirthi@gmail.com>

### DIFF
--- a/programs/pluto/ipsec_pluto.8.xml
+++ b/programs/pluto/ipsec_pluto.8.xml
@@ -685,7 +685,7 @@
     remap="B">whack</emphasis>.</para>
 
     <refsect2 id="ikes_job">
-      <title>IKE's Job</title>
+      <title>IKE&apos;s Job</title>
 
       <para>A <emphasis remap="I">Security Association</emphasis> (<emphasis
       remap="I">SA</emphasis>) is an agreement between two network nodes on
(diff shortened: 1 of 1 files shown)


letoams:
Can you show me when or where these are not correctly displayed?
I don't see any issues with the quotes you are fixinf.

tanuja3:
pluto

- At https://libreswan.org/man/ipsec_pluto.8.html the quotes are not being displayed correctly.

letoams:
On Wed, 15 Apr 2020, Tanuja Kirthi Doddapaneni wrote:

> Subject: Re: [libreswan/libreswan] Fixing manpage of pluto. (#314)

> pluto
>
>  *  At https://libreswan.org/man/ipsec_pluto.8.html the quotes are not being displayed correctly.

Ahh, I think that might be a bug in the xmlto package used to convert
the man pages, where in the man pages it shows properly.

I don't think single or double quotes should be replaced everyhwere, eg
see: https://stackoverflow.com/questions/24001878/do-i-need-to-replace-double-single-quote-in-xml-body-text

I'll try and regenerate the html and see if that bug has gone away yet.
It has been a while since we actually updated the html output.

Paul

tanuja3:
Oh okay, then I'll close this pull request. Thank you.

PR #321, opened 2020-03-27
pluto: add FIPS validation for EC keys for ECDSA

### LINKED ISSUE #318: Missing FIPS validation code for X.509 certificates with elliptic curve cryptography (ECC) keys
Hi,

I am currently running into a couple of issues with the following code in programs/pluto/connections.c:

```c
        if (libreswan_fipsmode()) {
                SECKEYPublicKey *pk = CERT_ExtractPublicKey(cert);
                passert(pk != NULL);
                if (pk->u.rsa.modulus.len * BITS_PER_BYTE < FIPS_MIN_RSA_KEY_SIZE) {
                        whack_log(RC_FATAL,
                                "FIPS: Rejecting cert with key size %d which is under %d",
                                pk->u.rsa.modulus.len * BITS_PER_BYTE,
                                FIPS_MIN_RSA_KEY_SIZE);
                        SECKEY_DestroyPublicKey(pk);
                        CERT_DestroyCertificate(cert);
                        return false;
                }
                /* TODO FORCE MINIMUM SIZE ECDSA KEY */
                SECKEY_DestroyPublicKey(pk);
        }
```
The two issues are:

1. When I use a X.509 certificate with elliptic curve (EC) keys, the code for RSA key size check above incorrectly runs which expectedly results in a failure.
2. As the TODO indicates, the code for FIPS validation of EC key is missing.

In order to resolve these issues, I am proposing adding code along the following lines:

```c
/* Get a pointer to the elliptic curve parameters that are binary encoded using DER. */
SECItem *curve_oid = &pk->u.ec.DEREncodedParams;
/* Get the tag enum assigned to the particular elliptic curve parameter set. */
SECOidTag curve_tag = SECOID_FindOIDTag(curve_oid);
/* Check if `curve_tag` matches any of the curves allowed by FIPS. */
switch (curve_tag) {
  case <any of FIPS certified curves, e.g. SEC_OID_SECG_EC_SECP384R1>:
    break;
  default:
    return false;
}
```

Is this the right approach to solving the aforementioned issues? Or is there a better alternative?

Thanks.

kavinda-forcepoint:
This change closes #318.

Prior to this change, there was no FIPS check for elliptic curve
cryptography (ECC) key curves used for ECDSA. So what happened was that
FIPS verification for RSA keys would run when EC keys were used which
resulted in failures.

The FIPS check logic for an EC curve in this change is identical to what
`certutil` does. That is, the logic uses Network Security Services (NSS)
to determine what curves are allowed in FIPS mode. NSS, meanwhile,
relies on the operating system security policy to know what is allowed.

An alternative implementation to check FIPS approval for an EC curve is
to explicitly check against the named curves listed in Appendix D in
https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf I decided
against this approach because it would require editing the code if the
approved set of curves changes in the future. It is better instead to let
the OS security policy that configures NSS to dynamically specify what is
allowed.


kavinda-forcepoint:
Hi @letoams,

Thanks for the update. Could you point me at the change where the fix was made? I would like to integrate that change into my Libreswan version.

letoams:
you need two commits:
https://github.com/libreswan/libreswan/commit/ad73bc5a5b2ea0d509feff07b56fbbdfe046a218
and
https://github.com/libreswan/libreswan/commit/101f48526d1848cc3b185b9617d19f1b614bc72e
For testing, there is a new test case https://github.com/libreswan/libreswan/commit/6cd7ee47d78f4084ce4e156f4f2acc1437da6863

kavinda-forcepoint:
Thank you for the quick reply, Paul. I am closing this PR as requested.

letoams:
note that the only EC curve not allowed by NIST is the p192 curve, and since that curve is not implemented by NSS, libreswan does not have to do any kind of checks for ECDSA.

PR #323, opened 2020-04-01
seamless nss cert rotation for existing tunnels

myungjin:
Libreswan does not support seamless cert rotation. Therefore, when
a new cert is issued and inserted into NSS DB, a tunnel should be
torn down and set up again or pluto needs to be restarted in order
to pick up the new cert. This causes temporary disruption for an
existing tunnel. This patch enables a new feature that allows
rotation of cert without bringing down any active tunnel by
replacing an old cert with a new one in pluto.

### DIFF
--- a/programs/pluto/cert_rotation.c
+++ b/programs/pluto/cert_rotation.c
@@ -1,0 +1,96 @@
+#include <nss.h>
+
+#include "cert_rotation.h"
+#include "connections.h"
+#include "constants.h"
+#include "defs.h"
+#include "keys.h"
+#include "log.h"
+#include "nss_cert_load.h"
+#include "whack.h"
+
+static inline cert_t backup_cert(struct end *dst)
+{
+	cert_t backup;
+
+	backup.ty = dst->cert.ty;
+	backup.u.nss_cert = dst->cert.u.nss_cert;
+
+	return backup;
+}
+
+static inline void restore_cert(struct end *dst, cert_t cert)
+{
+	// restore old certs
+	dst->cert.ty = cert.ty;
+	dst->cert.u.nss_cert = cert.u.nss_cert;
+}
+
+static inline void release_cert(cert_t cert)
+{
+	if (cert.u.nss_cert != NULL)
+		CERT_DestroyCertificate(cert.u.nss_cert);
+}
+
+// This function attempts to load a new cert if NSS DB has a new one
+static bool attempt_new_cert(struct fd *whackfd, struct end *dst)
+{
+	const char *nickname;
+	enum whack_pubkey_type pubkey_type;
+
+	nickname = cert_nickname(&dst->cert);
+	if (nickname == NULL) {
+		whack_comment(whackfd, "No cert nickname found");
+		return false;
+	}
+
+	pubkey_type = WHACK_PUBKEY_CERTIFICATE_NICKNAME;
+	if (!load_end_cert_and_preload_secret(whackfd, dst->leftright,
+					      nickname, pubkey_type, dst)) {
+		whack_comment(whackfd, "New cert attempt failed for %s", nickname);
+		return false;
+	}
+
+	return true;
+}
+
+static void rotate_end_cert(struct fd *whackfd, struct end *dst)
+{
+	cert_t old_cert;
+
+	if (dst->cert.u.nss_cert == NULL)
+		whack_comment(whackfd, "No cert exists for %s; nothing to do",
+			      dst->leftright);
+	else {
+		old_cert = backup_cert(dst);
+
+		if (!attempt_new_cert(whackfd, dst)) {
+			whack_comment(whackfd, "Cert rotation failed for %s",
+				      dst->leftright);
+			restore_cert(dst, old_cert);
+			return;
+		}
+
+		whack_comment(whackfd, "Cert rotation complete for %s",
+			      dst->leftright);
+		release_cert(old_cert);
+	}
+}
+
+void rotate_cert(struct fd *whackfd, const struct whack_message *wm)
+{
+	struct connection *conn;
+	struct end *dst;
+
+	conn = conn_by_name(wm->name, false);
+	if (conn == NULL)  {
+		whack_comment(whackfd, "No connection '%s' found", wm->name);
+		return;
+	}
+
+	dst = &conn->spd.this;
+	rotate_end_cert(whackfd, dst);
+
+	dst = &conn->spd.that;
+	rotate_end_cert(whackfd, dst);
+}
--- a/include/whack.h
+++ b/include/whack.h
@@ -333,6 +333,9 @@ struct whack_message {
 	/* for WHACK_REREAD */
 	u_char whack_reread;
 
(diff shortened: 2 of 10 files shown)


programs/pluto/cert_rotation.c
```diff
@@ -0,0 +1,50 @@
...
+
+	dst = &conn->spd.this;
+	nickname = cert_nickname(&dst->cert);
+	if (nickname == NULL) {
+		whack_comment(whackfd, "No cert nickname found");
+		return;
+	}
+
+	// temporarily back up cert to release or restore it later
+	old_nss_cert = dst->cert.u.nss_cert;
```
cagney, line 35: This is relying on load_end_cert_and_preload_secret() being hardwired to:
	dst_end->cert.ty = CERT_NONE;
	dst_end->cert.u.nss_cert = NULL;
I'd instead save the entire end->cert and clear it; and then restore the contents if things fail.
I'd make that explicit here, and remember to restore everything.  And remember to restore

programs/pluto/cert_rotation.c
```diff
@@ -0,0 +1,50 @@
...
+
+	pubkey_type = WHACK_PUBKEY_CERTIFICATE_NICKNAME;
+	if (!load_left_cert_and_secret(whackfd, nickname, pubkey_type, dst)) {
```
cagney, line 38: What happens if the connection isn't oriented, or itis "right" that is local.

cagney:
What happens if, instead, once authentication has completed the cert/secret are released?

programs/pluto/cert_rotation.c
myungjin: good point..  I will fix it.

programs/pluto/cert_rotation.c
myungjin: This simply loads latest cert and key from NSS DB and put them in pluto's memory.

This doesn't make any changes on an existing connection. The only assumption made here is that the connection is loaded into pluto.

I may be wrong, but I think as far as rotating cert is concerned, it doesn't matter whether the connection is up/down or what is the right end.

myungjin:
@cagney I don't understand your concern on cert/secret release issue. If I am not wrong, what I observed so far is that cert and secret are always available in pluto's data structure as long as the connection is loaded. Please correct me if I miss something.

myungjin:
@sebspion What would be the next step? As this PR grows old, it gets conflicts. Should I keep resolving conflicts until it gets merged?

bleve:
Ignore sebspion. He is just a random user.

letoams:
On Fri, 17 Apr 2020, myungjin wrote:

> @sebspion What would be the next step? As this PR grows old, it gets conflicts. Should I keep resolving conflicts until it gets merged?

Thanks for the followup. We do want a solution for this problem but
I don't think we agree on the solution being correct.

Andrew had some concerns that I don't think are addressed yet?

Paul

myungjin:
@letoams @cagney Thank you for your reply. I am willing to continue to make changes, but Libreswan codebase is quite huge, and I may not have clear understanding on Andrew's concerns.

Can/Will cert/secret be released after authentication?

I was only able to find that the cert associated with connection is released at  https://github.com/libreswan/libreswan/blob/master/programs/pluto/connections.c#L166. It looks like that delete_end() is called when either connection is terminated (i.e., delete_connection()) or adding connection is failed (i.e., add_connection()).

So please let me know what part I should look into to address your concerns.

myungjin:
@letoams @cagney Any suggestion on this PR?

letoams:
On Mon, 18 May 2020, myungjin wrote:

> @letoams @cagney Any suggestion on this PR?

Maybe we should schedule a conference call and talk about this. It would
be good to resolve this issue properly.

Can you email me your contact info? I'm paul@nohats.ca

Also what time next week would work for you? In which timezone? Both
cagney and I are in Eastern.

Paul

cagney:
I've pushed f957cd0206a2580c3ce2e4feb7ee83ad03ce2429
It will let you eliminate load_left_cert_and_secret()

programs/pluto/cert_rotation.c
cagney: > This simply loads latest cert and key from NSS DB and put them in pluto's memory.
> 
> This doesn't make any changes on an existing connection. The only assumption made here is that the connection is loaded into pluto.

It changes the existing connection's cert, it also potentially changes the .id, and possibly needs to update .ca (for instance when the original .ca was extracted from the cert, current code doesn't handle this).






> 
> I may be wrong, but I think as far as rotating cert is concerned, it doesn't matter whether the connection is up/down or what is the right end.

At any time a connection is either oriented (left/right somehow map onto that/this) or unoriented (this/that) are arbitrary.  The patch assumes this==left and only end's cert needs an update.  There's nothing to say that .that is the end needing an update.

cagney:
> @cagney I don't understand your concern on cert/secret release issue. If I am not wrong, what I observed so far is that cert and secret are always available in pluto's data structure as long as the connection is loaded. Please correct me if I miss something.

I'm just asking a question.  What would happen if the cert was released after things were authenticated?

I'm sure we don't want to release the secret - loading it is expensive.  I'm not so sure about the cert.

letoams:
On Tue, 26 May 2020, cagney wrote:

>       @cagney I don't understand your concern on cert/secret release issue. If I am not wrong, what I observed so far is
>       that cert and secret are always available in pluto's data structure as long as the connection is loaded. Please
>       correct me if I miss something.
> 
> I'm just asking a question. What would happen if the cert was released after things were authenticated?
> 
> I'm sure we don't want to release the secret - loading it is expensive. I'm not so sure about the cert.

an atomic update of the pointer from the old cert to the new cert would
be nice. The secret should not change. Although if the certificate
is replaced including a new private key, then I'm fine with restarting
all connections using it.

Similarly, if we update the cert of the peer, when using hardcoded certs
in nss instead of using a CA, I'm also fine with just reloading the
connection(s).

The real use case is the remote access VPN or Opportunistic IPsec, where
we have many connections, but the only thing we really use our own cert for
(After the connection is loaded) is creating the CERT payload. Perhaps
we should make this explicit and build the CERT payload for later
re-use?

Paul

myungjin:
> I've pushed f957cd0
> It will let you eliminate load_left_cert_and_secret()

@cagney Thank you for the change. One issue with using load_end_cert_and_preload_secret is that the function is static; it is not visible from other source files. Is it okay to change its signature and use extern directive?

myungjin:
@cagney I took freedom to remove static keyword in front of load_end_cert_and_preload_secret(). If it's not desirable, let me know. I will revert it.

myungjin:
@letoams @cagney I refactored the PR as we discussed.

One thing to note is that cert rotation of `left` and `right` ends are not transactional.
That is, even though the cert rotation fails for one of them, there is no rollback.
I think this is okay because each cert should be able to updated independent of other certs.
Otherwise, please let me know.

letoams:
merged. Thanks again for the original patch !

myungjin:
@letoams Thank you, Paul!

PR #336, opened 2020-05-10
documentation: remove redundant options in whack --help output

vukasink:
removed stuff appears couple of rows below


letoams:
I almost rejected the patch at the the whack command has [options] --to [options] where these can repeat, but the --help indeed only lists --to at the end and doesn't repeat the set, so your patch is correct. Merged.

PR #338, opened 2020-05-15
ddos-cookies-threshold and ddos-max-halfopen parameters are swapped

johnmah:
the settings for ddos-cookies-threshold and ddos-max-halfopen are swapped in `ipsec whack --status` output.


letoams:
merged in, thanks!

PR #345, opened 2020-05-23
Use pkg-config recommended by the environment

dkg:
This should help with cross-building.  See suggestions from Helmut
Grohne on https://bugs.debian.org/958355

Signed-off-by: Daniel Kahn Gillmor <dkg@fifthhorseman.net>


bleve:
This change is not to correct location. setting for pkg-config belongs to mk/config.mk

dkg:
I'd be fine with any fix that you think is better -- feel free to change it to whatever you think is correct, as long as it makes it easier to cross-compile by indicating the correct pkg-config to use, based on a PKG_CONFIG setting.

bleve:
Fixed with slightly different change.

cagney:
I tweaked this (and followed up by ripping out a pile of crud).
Please test.

PR #350, opened 2020-07-02
pluto: fix issues with wildcards in peer DNs of connection configs.

kavinda-forcepoint:
This PR resolves the following issues:

 1. There were several bugs preventing IKE connection establishment when
    specifying wildcards in the peer DN (Distinguished Name) via `leftid`
    and/or `rightid` for either or both tunnel endpoints.

 2. Specifying a wildcard in a peer DN marked a connection config as a template
    that would not be instantiated _even_ when there was sufficient information
    to initiate an IKE connection in the config.

Fixing issue #1 means:

* You can now specify a wildcard for the entire value of a Relative
  Distinguished Name (RDN) in a peer DN.
  ** E.g.: `leftid="CN=test.localdomain, O=*"`
* You can specify wildcards for multiple RDNs in a peer DN.
  ** E.g.: `rightid="CN=*, O=*"`

Fixing issue #2 means:

* If an IKE connection config is a template _only_ because the peer DN has a
  wildcard _and_ the connection config has sufficient information to initiate
  the connection, `pluto` instantiates a template instance and initiates an IKE
  connection. `pluto` does this automatically for connections configured with
  `auto-start`.

I have successfully tested this commit manually for a IKE tunnel configured with:

* FIPS enabled
* ECDSA for authentication
* Wildcards in peer DNs


kavinda-forcepoint:
Hi @letoams,

Do you have any further comments on this PR?

Thanks.

-Kavinda

letoams:
I'm a bit confused about this issue.

If you specify an RDN with wildcards, doesn't that mean that multiple clients with different RDN's are expected to connect. So then instantiating seems the right thing to do ?

If there is only one possible remote peer, why are you using wildcards in the peer id= ?

kavinda-forcepoint:
> If you specify an RDN with wildcards, doesn't that mean that multiple clients with different RDN's are expected to connect.

That is one of the use cases, i.e. we expect multiple clients with different RDNs to connect.

> So then instantiating seems the right thing to do ?

Correct. My changes result in the template getting instantiated. Prior to my changes, that didn't happen.

> If there is only one possible remote peer, why are you using wildcards in the peer id= ?

We need support for a scenario where the peer needs to only match a sufficient number of RDNs in the peer DN.

kavinda-forcepoint:
This issue has been resolved. Wildcards in peer DNs work correctly now.

PR #355, opened 2020-08-05
Declare the build dependency on nss-tools

skitt:
The Fedora .spec file relies on certutil, and that is provided by the
nss-tools package; this makes sure the latter is installed.

Signed-off-by: Stephen Kitt <skitt@redhat.com>


letoams:
merged. thanks!

PR #357, opened 2020-08-08
Fix labeled IPsec

bauen1:
It appears that 998aa5f18bad901d6f5c834494d06ee405053b59 changed
policy_label from being an "option" to a "string", but forgot to change
the conditional for (optionally) sending the value to pluto.
Using 3.32 this results in addconn parsing the configuration file but
simply **never** sends this value to pluto.
This breaks Labeled IPSec support.

This was only tested as a patch against the debian package for 3.32-3


letoams:
merged, thanks. Do you have a full name for the credits or is "bauen1" okay for you ? :)

bauen1:
I use bauen1 online, so that is fine, if you need a real name for legal purposes: Jonathan Hettwer

PR #360, opened 2020-09-02
add missing include of limits.h for UINT_MAX

maxice8:
note: 'UINT_MAX' is defined in header '<limits.h>'; did you forget to '#include <limits.h>'?
   92 | # include "kernel_xfrm_interface.h"
  +++ |+#include <limits.h>

Fixes building on musl


letoams:
merged. thanks!

PR #361, opened 2020-09-02
testing: Updated NFS mount details of OpenBSD

RaviTejaCMS:
- Moved the NFS Mount to swandefault network
- Added swandefault network to openbsd(east and west) definition files
- Updated comment's
- Updated iked_flags


letoams:
merged. thanks!

PR #382, opened 2020-11-07
pluto: add initial labeled IPsec Traffic Selector support for IKEv2

kavinda-forcepoint:
- The implementation is based on the RFC draft at
  https://tools.ietf.org/html/draft-ietf-ipsecme-labeled-ipsec-04

- This initial implementation does NOT support the following:
  - Optional labeling
  - Multiple labels for in a single Traffic Selector payload
  - Different security labels for TSi/TSr
  - Different security labels for different IP ranges

- The above means that if security labels are used, then they are required on
  _both_ sides of a tunnel.


lib/libswan/constants.c
```diff
@@ -1833,17 +1833,35 @@ enum_names ikev2_notify_names = {
...
+
+/**
+ * ikev2_ts_type_name_private_use: Traffic Selector Types allocated for private use.
+ *
+ * The reason this is required is that private Traffic Selector Type values are
+ * not contiguous with publicly allocated TS Type values.
+ */
+static const char *const ikev2_ts_type_name_private_use[] = {
+	"IKEv2_TS_SECLABEL",		/* 241 */
+};
```
sahanaprasad07, line 1849: insert empty line after this

programs/pluto/ikev2_ts.h
```diff
@@ -33,6 +33,20 @@ struct traffic_selector {
...
+
+	/**
+	 * `sec_ctx` is the security label (if any) that is associated with
+	 * this Traffic Selector.
+	 * 
+	 * There are 2 cases where `sec_ctx` is non-NULL:
+	 *
+	 *  1. If `ts_type` == `IKEv2_TS_SECLABEL`
+	 *
+	 *  2. If `ts_type` corresponds to an address range _and_ the security
```
sahanaprasad07, line 45: shouldn't it always be the second case? i.e TS_SECLABEL should always be in conjunction with other parameters (like the address range that is mandatory)

programs/pluto/ikev2_parent.c
```diff
@@ -4215,7 +4228,82 @@ static bool ikev2_rekey_child_resp(struct ike_sa *ike, struct child_sa *child,
...
+ *
+ * @param[in,out]	new_state	State of the new child/IPsec SA.
+ * @param[in]		predecessor	State of the predecessor child/IPsec SA.
+ *
+ * @return	True if copying the security label from `predecessor` to
+ * 		`new_state` was successful.
+ */
+static bool ikev2_rekey_child_copy_sec_ctx(struct state *const new_state,
+					   struct state const* const predecessor) {
+	libreswan_log("%s: Prior to updating new SA's security label: predecessor security label = %s, new child/IPsec SA security label = %s",
```
sahanaprasad07, line 4244: would be nice to split this log line into to lines.

programs/pluto/ikev2_parent.c
```diff
@@ -4215,7 +4228,82 @@ static bool ikev2_rekey_child_resp(struct ike_sa *ike, struct child_sa *child,
...
+			if (predecessor->sec_ctx == new_state->sec_ctx) {
+				return true;
+			}
+			if (streq(predecessor->sec_ctx->sec_ctx_value,
+				  new_state->sec_ctx->sec_ctx_value)) {
+				return true;
+			}
+
+			/* Security label mismatch between old and new child/IPsec SAs.*/
+			loglog(RC_LOG_SERIOUS, "ERROR: Security label mismatch between old SA (%s) and new SA (%s) during child rekeying", (predecessor->sec_ctx ? predecessor->sec_ctx->sec_ctx_value: "<NULL>"), (new_state->sec_ctx ? new_state->sec_ctx->sec_ctx_value : "<NULL>"));
```
sahanaprasad07, line 4265: would be nice to split this log line into to lines.

programs/pluto/ikev2_parent.c
```diff
@@ -4215,7 +4228,82 @@ static bool ikev2_rekey_child_resp(struct ike_sa *ike, struct child_sa *child,
...
+		}
+	} else {
+		/* Predecessor does NOT have a security label. */
+		if (new_state->sec_ctx != NULL) {
+			/*
+			 * New child/IPsec SA has a security label while the
+			 * predecessor doesn't. This should not happen -
+			 * possible bug.
+			 */
+			loglog(RC_LOG_SERIOUS, "ERROR: Security label mismatch between old SA (<NULL>) and new SA (%s) during child rekeying", new_state->sec_ctx->sec_ctx_value);
```
sahanaprasad07, line 4284: would be nice to split this log line into to lines.

programs/pluto/ikev2_ts.c
```diff
@@ -125,13 +137,58 @@ static stf_status ikev2_emit_ts(pb_stream *outpbs,
...
 		if (!out_struct(&its, ts_desc, outpbs, &ts_pbs))
 			return STF_INTERNAL_ERROR;
 	}
 
+	/*
+	 * Output a TS_SECLABEL substructure as part of the TS Payload if a
+	 * security label exists.
+	 */
+	if (ts->sec_ctx) {
+		/* Sanity check: +1 is for the NUL-termination character. */
```
sahanaprasad07, line 160: NULL

programs/pluto/ikev2_ts.c
```diff
@@ -125,13 +137,58 @@ static stf_status ikev2_emit_ts(pb_stream *outpbs,
...
+				ts->sec_ctx->sec_ctx_value);
+			return STF_INTERNAL_ERROR;
+		}
+
+		/* Initialize the header of the TS_SECLABEL substructure payload. */
+		struct ikev2_ts_seclabel ts_seclabel;
+		ts_seclabel.isa_tssec_type = IKEv2_TS_SECLABEL;
+		ts_seclabel.isa_tssec_reserved = 0;
+		/* Length of the TS_SECLABEL substructure = 4 (size of header) + security label length */
+		ts_seclabel.isa_tssec_sellen = 4 + ts->sec_ctx->ctx.ctx_len;
```
sahanaprasad07, line 174: it would be nice to use a macro for 4

programs/pluto/ikev2_ts.c
```diff
@@ -320,6 +379,111 @@ stf_status v2_emit_ts_payloads(const struct child_sa *child,
...
+	incoming_ctx.ctx.ctx_len = labelLen;
+
+	/* Parse the security label value of the TS_SECLABEL substructure. */
+	if (!in_raw(incoming_ctx.sec_ctx_value, labelLen, ins,
+				"Traffic Selector Security Label")) {
+		loglog(RC_LOG_SERIOUS, "ERROR: Could not read TS_SECLABEL security label value");
+		return false;
+	}
+
+	/* Incoming security label MUST be NUL-terminated. */
```
sahanaprasad07, line 469: NULL

programs/pluto/ikev2_ts.c
```diff
@@ -649,26 +900,502 @@ static int score_address_range(const struct end *end,
 	return f;
 }
 
+/**
+ * seclabel_narrow_status: Indicates the outcome of a security label "narrowing."
```
sahanaprasad07, line 904: The draft mentions that security label should not be narrowed.

   The mechanism of narrowing of Traffic Selectors with
   TS_IPV4_ADDR_RANGE and TS_IPV6_ADDR_RANGE does not apply to
   TS_SECLABEL as the Security Label itself is not interpreted and
   cannot itself be narrowed

programs/pluto/ikev2_ts.c
```diff
@@ -340,13 +504,55 @@ static bool v2_parse_ts(struct payload_digest *const ts_pd,
...
+		uint8_t ts_type;
+		/*
+		 * Peek at the Traffic Selector Type in the input stream in order to figure
+		 * out which substructure to process next.
+		 */
+		if (!peek_raw(&ts_type, sizeof(ts_type), &ts_pd->pbs, "Traffic Selector Type")) {
+			return false;
+		}
+
+		if (ts_type == IKEv2_TS_SECLABEL) {
```
sahanaprasad07, line 541: this could have been checked as an additional case in the switch below (along with v4 and v6 traffic types)

lib/libswan/constants.c
kavinda-forcepoint: Fixed.

programs/pluto/ikev2_ts.h
kavinda-forcepoint: The first case is used during input TS Payload processing. I have added several comments to clarify this. Please let me know if you think it needs more explanation.

programs/pluto/ikev2_parent.c
kavinda-forcepoint: Fixed.

programs/pluto/ikev2_parent.c
kavinda-forcepoint: Fixed.

programs/pluto/ikev2_parent.c
kavinda-forcepoint: Fixed.

programs/pluto/ikev2_ts.c
kavinda-forcepoint: I meant it as the NUL character: https://en.wikipedia.org/wiki/Null_character

programs/pluto/ikev2_ts.c
kavinda-forcepoint: Good idea!

programs/pluto/ikev2_ts.c
kavinda-forcepoint: Same as above regarding the NUL character. But let me know if you would like me to change it to NULL still though.

programs/pluto/ikev2_ts.c
kavinda-forcepoint: You are correct. I used "narrow" because I couldn't think of a better word while implementing this logic; hence, why I used quotes around it.

Do you think something like "select" would be better, i.e. "Indicates the outcome of a security label selection."

If you have a better term, please let me know!

programs/pluto/ikev2_ts.c
kavinda-forcepoint: I did have it as an additional case in the `switch` initially, but that resulted in duplicating calls to `in_struct()` for `&ts1` and `++addr_range_count`. What I ended up having was the cleanest code I could write.

But let me know if I should still add a case to the `switch`.

PR #389, opened 2021-01-08
Clang build fixes

tbaederr:
See individual commits

### DIFF
--- a/lib/libswan/alloc_printf.c
+++ b/lib/libswan/alloc_printf.c
@@ -20,6 +20,7 @@
 
 #include "lswalloc.h"
 
+__attribute__((__format__(__printf__, 1, 0)))
 char *alloc_vprintf(const char *fmt, va_list master_ap /* must va_copy */)
 {
 	int length;
--- a/lib/libswan/jam_nss_cka.c
+++ b/lib/libswan/jam_nss_cka.c
@@ -24,7 +24,7 @@
 size_t jam_nss_cka(struct jambuf *buf, CK_ATTRIBUTE_TYPE attribute)
 {
 	switch (attribute) {
-#define CASE(T) case T: return jam_string(buf, #T + strlen("CKA_"))
+#define CASE(T) case T: return jam_string(buf, &#T[strlen("CKA_")])
 		CASE(CKA_DERIVE);
 		CASE(CKA_FLAGS_ONLY);
 		CASE(CKA_WRAP);
--- a/lib/libswan/jam_nss_ckf.c
+++ b/lib/libswan/jam_nss_ckf.c
@@ -29,7 +29,7 @@ size_t jam_nss_ckf(struct jambuf *buf, CK_FLAGS flags)
 #define FLAG(F) \
 	if (flags & F) {					\
 		size += jam_string(buf, sep);			\
-		size += jam_string(buf, #F + strlen("CKF_"));	\
+		size += jam_string(buf, &#F[strlen("CKF_")]);	\
 		sep = "+";					\
 		flags ^= F;					\
 	}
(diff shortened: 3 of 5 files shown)


cagney (changes requested):
- header lswcdefs.h defines the macro:
`# define PRINTF_LIKE(n) __attribute__ ((format(printf, n, n + 1)))`
can an equivalent VPRINTF_LIKE() macro be defined?
- would it be better to attach this to the declaration - that way the compiler can also check callers?

tbaederr:
At least `alloc_vprintf` already has a `PRINTF_LIKE` attribute in the declaration, but I guess that's not being used when compiled with clang because `GCC_LINT` is not set.

cagney (commented):
The idiom exists for a reason.
clang made their error non-specific: https://reviews.llvm.org/D55382
gcc doesn't support this:
https://gcc.gnu.org/pipermail/gcc-patches/2017-February/469872.html
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62181

cagney:
> At least `alloc_vprintf` already has a `PRINTF_LIKE` attribute in the declaration, but I guess that's not being used when compiled with clang because `GCC_LINT` is not set.

 `mk/config.mk` has:
```
ifeq ($(origin GCC_LINT),undefined)
GCC_LINT = -DGCC_LINT
endif
USERLAND_CFLAGS += $(GCC_LINT)
```
so GCC_LINT should be defined?

Can I also suggest splitting these two patches, they are for different problems.

tbaederr:
> The idiom exists for a reason.
> clang made their error non-specific: https://reviews.llvm.org/D55382
> gcc doesn't support this:
> https://gcc.gnu.org/pipermail/gcc-patches/2017-February/469872.html
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62181

I know, I'm not a fan of the warning either. But if libreswan should be buildable with clang and you don't want to have clang-specific cflags somewhere, this is how to fix it I guess.

tbaederr:
> > At least `alloc_vprintf` already has a `PRINTF_LIKE` attribute in the declaration, but I guess that's not being used when compiled with clang because `GCC_LINT` is not set.
> 
> `mk/config.mk` has:
> 
> ```
> ifeq ($(origin GCC_LINT),undefined)
> GCC_LINT = -DGCC_LINT
> endif
> USERLAND_CFLAGS += $(GCC_LINT)
> ```
> 
> so GCC_LINT should be defined?
> 
> Can I also suggest splitting these two patches, they are for different problems.

I'll have to investigate. Split them how into what? Two PRs?

cagney:
> > The idiom exists for a reason.
> > clang made their error non-specific: https://reviews.llvm.org/D55382
> > gcc doesn't support this:
> > https://gcc.gnu.org/pipermail/gcc-patches/2017-February/469872.html
> > https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62181
> 
> I know, I'm not a fan of the warning either. But if libreswan should be buildable with clang and you don't want to have clang-specific cflags somewhere, this is how to fix it I guess.

Yes.  Might have to just suck it up.
I've pointed DHR at this on IRC.

cagney:
> > > At least `alloc_vprintf` already has a `PRINTF_LIKE` attribute in the declaration, but I guess that's not being used when compiled with clang because `GCC_LINT` is not set.
> > 
> > 
> > `mk/config.mk` has:
> > ```
> > ifeq ($(origin GCC_LINT),undefined)
> > GCC_LINT = -DGCC_LINT
> > endif
> > USERLAND_CFLAGS += $(GCC_LINT)
> > ```
> > 
> > 
> > so GCC_LINT should be defined?
> > Can I also suggest splitting these two patches, they are for different problems.
> 
> I'll have to investigate. Split them how into what? Two PRs?

Branch?  New pull request?  (I really don't know how to drive github)

For GCC_LINT, let us know, it might be easier to drop the #ifdef GCC_LINT wrapper

tbaederr:
There are a bunch of other problems, but I'll create separate pull requests for all of them, starting with https://github.com/libreswan/libreswan/pull/393. Thanks.

PR #393, opened 2021-01-11
Add and use PRINTF_LIKE_VA()

tbaederr:
Clang is more strict about checking printf-like format strings. Add a
PRINTF_LIKE_VA() macro that evaluates to the proper __attribute__(())
and use that where needed.


cagney:
Nice.  There were more than I expected.
Paul, care to merge.

PR #402, opened 2021-01-19
pluto: Use NULL when refering to pointers

tbaederr:
The code block here defines an array of `char*` pointers, but then sets
the first element to `'\0'`, which is a char an not a pointer. Set it to
`NULL` instead.
This fixes the following warning from clang:

```
programs/pluto/ikev1_spdb_struct.c:1278:18: error: expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Werror,-Wnon-literal-null-conversion]
                                                nbyte[0] = '\0';
                                                           ^~~~
```

### DIFF
--- a/programs/pluto/ikev1_spdb_struct.c
+++ b/programs/pluto/ikev1_spdb_struct.c
@@ -1275,7 +1275,7 @@ bool ikev1_out_sa(pb_stream *outs,
 						};
 
 						char *nbyte[1];
-						nbyte[0] = '\0';
+						nbyte[0] = NULL;
 
 						if (!out_struct(&attr,
 								attr_desc,


letoams:
On Tue, 19 Jan 2021, tbaederr wrote:

> The code block here defines an array of char* pointers, but then sets
> the first element to '\0', which is a char an not a pointer. Set it to
> NULL instead.
> This fixes the following warning from clang:
> 
> programs/pluto/ikev1_spdb_struct.c:1278:18: error: expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Werror,-Wnon-li
> teral-null-conversion]
>                                                 nbyte[0] = '\0';
>                                                            ^~~~

This code has been removed already in a pending patch. It should hit the
repo in a day or so.

Paul

tbaederr:
Sure, fine with me. Thanks.

PR #403, opened 2021-01-19
pluto: Add missing assignment operators

tbaederr:
Omitting the assignment operator is a GNU extension which clang does not
support:
```
programs/pluto/state.c:116:21: error: use of GNU 'missing =' extension in designator [-Werror,-Wgnu-designator]
        [STATE_IKEv1_ROOF] &state_ikev1_roof,
                           ^
                           =
```


cagney (approved):
Bizarre.
Yes, definitely.
BTW, is there a way to hobble this feature when using gcc so it doesn't happen again?

tbaederr:
I'm not sure how to disable it off the top of my head. I thought switching from gnu99 to c99 would help but that just makes gcc complain about all the uses of `typeof`.

cagney:
does adding -Wgnu-designator to the warning flags help?

PR #404, opened 2021-01-20
testing: switch argument order of check_jambuf()

tbaederr:
The argument order in combination with how va_start was used on them
causes clang to print an error:

```
testing/check/jambuf/jambufcheck.c:85:16: error: passing an object that undergoes default argument promotion to 'va_start' has undefined behavior [-Werror,-Wvarargs]
                va_start(ap, ok);
                             ^
testing/check/jambuf/jambufcheck.c:26:51: note: parameter of type 'bool' is declared here
static void check_jambuf(const char *expect, bool ok, ...)
                                                  ^
```

### DIFF
--- a/testing/check/jambuf/jambufcheck.c
+++ b/testing/check/jambuf/jambufcheck.c
@@ -23,7 +23,7 @@
 
 unsigned fails;
 
-static void check_jambuf(const char *expect, bool ok, ...)
+static void check_jambuf(bool ok, const char *expect, ...)
 {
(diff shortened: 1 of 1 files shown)


cagney (approved):
Ok.
Code is compiled with strict-prototypes so both the caller and callee know the exact size of the parameter so, like for float vs double, no promotion should be needed; C disagrees :-(

letoams:
merged (but I forgot to reference this PR, so it won't auto close)

PR #406, opened 2021-01-28
Use close_range to close fds

skitt:
On some systems, the file descriptor table can be large (524,288
entries on the system I’m writing this on), and the fd-closing loop in
plutomain takes a long time (2-3 *minutes*).

Linux 5.10 introduced the close_range syscall which can close huge
ranges of fds very quickly. This patch enables pluto to use this, when
it’s available.

The Makefile snippet changes detect support for close_range at build
time, using linux/close_range.h as a proxy (if it’s present, the
syscall number will also be present). At runtime, the syscall is
attempted in all cases if pluto was built with it; older kernels will
fail with ENOSYS. In any case, if close_range fails, we fall back to
the existing fd-closing behaviour, which is preserved anyway because
we want to loop over the fds from 0 to ctl_fd (excluded).

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -604,6 +604,10 @@ USERLAND_CFLAGS += -DLIBCURL
 CURL_LDFLAGS ?= -lcurl
 endif
 
+ifeq ($(USE_CLOSE_RANGE),true)
+USERLAND_CFLAGS += -DUSE_CLOSE_RANGE
+endif
+
 # Build support for the Linux Audit system
 
 USE_LINUX_AUDIT ?= false
--- a/mk/defaults/linux.mk
+++ b/mk/defaults/linux.mk
@@ -8,6 +8,9 @@ USERLAND_CFLAGS += -Dlinux
 # Expose pipe2() which is always available on BSD, what else?
 USERLAND_CFLAGS += -D_GNU_SOURCE
 
+# close_range
+USE_CLOSE_RANGE := $(shell if echo | gcc -include linux/close_range.h -E - > /dev/null 2>&1; then echo true; else echo false; fi)
+
 PORTDEFINE=-DSCANDIR_HAS_CONST
 
 # Detect linux variants and releases.
--- a/programs/pluto/pluto_seccomp.c
+++ b/programs/pluto/pluto_seccomp.c
@@ -118,6 +118,12 @@ static void init_seccomp(uint32_t def_action, bool main, struct logger *logger)
 		LSW_SECCOMP_ADD(uname);
 		LSW_SECCOMP_ADD(unlink);
 		LSW_SECCOMP_ADD(unlinkat);
+#ifdef USE_CLOSE_RANGE
+#ifndef __SNR_close_range
+#define __SNR_close_range __NR_close_range
+#endif
+		LSW_SECCOMP_ADD(close_range);
+#endif
 	}
 
 	/* common to pluto and helpers */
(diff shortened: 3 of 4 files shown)


mangelajo:
+1 I'm suffering this issue in my system. Pluto takes 3 min to start on a i7 skylake .

mangelajo (approved):

letoams (changes requested):
If pluto can try the new syscall and fallback if it doesnt exist, could we not just always do it and avoid the build option USE_CLOSE_RANGE alltogether ?

We will also need to update the SECCOMP white list to include this syscall

letoams:
Oh, I guess for compile time we still need to know it.

Ok, I'll tweak it and do the seccomp support

skitt:
> Oh, I guess for compile time we still need to know it.

Yup, I was about to reply exactly that ;-).

> Ok, I'll tweak it and do the seccomp support

Cool, thanks!

skitt:
> We will also need to update the SECCOMP white list to include this syscall

That turned out to be easy enough, I went ahead and did it (I think).

cagney:
Why is the code needed at all?

cagney:
> Why is the code needed at all?

To be more specific.  Other than 0,1,2 lockfile, and the control socket, what file descriptors are open that need closing?

cagney:
Assuming this code is correct:
```
	fprintf(stderr, "checking for open FDs\n");
	for (int fd = getdtablesize() - 1; fd >= 0; fd--) {
		struct stat s;
		if (fstat(fd, &s) == 0) {
			fprintf(stderr, "%d open\n", fd);
		}
	}
```
there's only:

Jan 28 13:45:17 west pluto[746]: checking for open FDs
Jan 28 13:45:17 west pluto[746]: 4 open
Jan 28 13:45:17 west pluto[746]: 3 open
Jan 28 13:45:17 west pluto[746]: 2 open
Jan 28 13:45:17 west pluto[746]: 1 open
Jan 28 13:45:17 west pluto[746]: 0 open

which, to my point, means there's nothing to close - if there was it would surely break the library that opened it.

skitt:
> > Why is the code needed at all?
> 
> To be more specific. Other than 0,1,2 lockfile, and the control socket, what file descriptors are open that need closing?

None in most cases, but more generally, any that are inherited from the parent process, and any that end up opened by future changes.

cagney:
> None in most cases, but more generally, any that are inherited from the parent process, and any that end up opened by future changes.

I pushed 5d9bd100e85181a68a8ee3e5ab3a7e8e9e2e403a
The code's gone.

skitt:
> > None in most cases, but more generally, any that are inherited from the parent process, and any that end up opened by future changes.
> 
> I pushed [5d9bd10](https://github.com/libreswan/libreswan/commit/5d9bd100e85181a68a8ee3e5ab3a7e8e9e2e403a)
> The code's gone.

Fair enough, I can’t think of a scenario where leaking fds would be bad for `pluto` itself.

cagney:
> Fair enough, I can’t think of a scenario where leaking fds would be bad for `pluto` itself.

FYI, testing turned up a scenario where leaking inherited FDs was good for pluto.  libfaketime uses FDs to keep time in sync between processes but Pluto was closing them :-(  See c6c410d1f004b5afc36c284628d15fa1fae2170d.

PR #412, opened 2021-02-04
fix more minor spelling issues

dkg:


programs/pluto/ikev2.c
```diff
@@ -1214,7 +1214,7 @@ static bool is_duplicate_request(struct ike_sa *ike,
 		return true;
 	}
 
-	struct v2_incomming_fragments *frags = ike->sa.st_v2_incomming[MESSAGE_REQUEST];
+	struct v2_incoming_fragments *frags = ike->sa.st_v2_incoming[MESSAGE_REQUEST];
```
cagney, line 1217: this breaks the build

dkg: is there some other "incomming" that i failed to change?  the build seems to work fine with a `s/incomming/incoming/g` fix applied everywhere i could find it, but maybe i somehow missed something?

I see d00e2003f7597667cd6e0047c3cb7bc1e3c38af1 appears to do a bit more cleanup.

PR #417, opened 2021-02-17
whack: global-redirect and global-redirect-to options

pietro:
We have IPsec gateways running Libreswan on cloud infrastructure. We have successfully used the config options `global-redirect` and `global-redirect-to` so Libreswan acts as a Load Balancer. Due to the changing nature of the cloud we have an external process that checks the status and IP address of our servers. We'd like to be able to change the redirect targets on the fly without having to restart pluto.

This patch adds the following argumenbts to `ipsec whack`:
- `--global-redirect` with valid options yes, no and auto
- `--global-redirect-to` comma separated list or global redirect targets


letoams:
merged into 4.3 (but forgot to reference this PR)

PR #418, opened 2021-02-23
pluto: introducing stats for redirection failures and completions

cliveza:
Reworked based on comments.


letoams:
merged in, thanks!

PR #419, opened 2021-02-24
pluto: fix IKEv2 labeled IPsec issues at Responder

kavinda-forcepoint:
* Use the TS_SECLABEL security label arriving from the Initiator for the
  child/IPsec SA instead of the `policy-label` from the connection
  configuration.
*  Fix bug where an extra SPD entry was created with the incorrect security label.
   * SPD entries should only have labels specified in `policy-label` of connection configurations.

### DIFF
--- a/programs/pluto/ikev2_parent.c
+++ b/programs/pluto/ikev2_parent.c
@@ -5943,8 +5943,6 @@ void ikev2_initiate_child_sa(struct pending *p)
 			 * from a policy we gave the kernel, so it _should_ be within our range?
 			 */
 			child->sa.st_acquired_sec_label = clone_hunk(p->sec_label, "st_acquired_sec_label");
-			c->spd.this.sec_label = clone_hunk(p->sec_label, "updated conn label");
-			c->spd.that.sec_label = clone_hunk(p->sec_label, "updated conn label");
 		}
 
 	} else {
--- a/programs/pluto/ikev2_ts.c
+++ b/programs/pluto/ikev2_ts.c
@@ -1065,16 +1065,14 @@ bool v2_process_ts_request(struct child_sa *child,
 
 		chunk_t const *const selected_sec_label =
 			score_ends_seclabel(&ends, c, &tsi, &tsr, child->sa.st_logger);
-		if (is_seclabel_required) {
-			if (!selected_sec_label) {
-				/* Security label required, but not found. */
-				continue;
-			}
-		} else {
-			if (selected_sec_label) {
-				/* Security label *not* required, but found. */
-				continue;
-			}
+		if (is_seclabel_required != (selected_sec_label != NULL)) {
+			/*
+			 * Either:
+			 *  - Security label required, but not found.
+			 *    OR
+			 *  - Security label *not* required, but found.
+			 */
+			continue;
 		}
 
 		enum fit responder_fit =
@@ -1178,16 +1176,14 @@ bool v2_process_ts_request(struct child_sa *child,
 
 				chunk_t const *const selected_sec_label =
 					score_ends_seclabel(&ends, d, &tsi, &tsr, child->sa.st_logger);
-				if (is_seclabel_required) {
-					if (!selected_sec_label) {
-						/* Security label required, but not found. */
-						continue;
-					}
-				} else {
-					if (selected_sec_label) {
-						/* Security label *not* required, but found. */
-						continue;
-					}
+				if (is_seclabel_required != (selected_sec_label != NULL)) {
+					/*
+					 * Either:
+					 *  - Security label required, but not found.
+					 *    OR
+					 *  - Security label *not* required, but found.
+					 */
+					continue;
 				}
 
 				struct best_score score = score_ends_iprange(responder_fit, d/*note D*/,
@@ -1476,16 +1472,14 @@ bool v2_process_ts_response(struct child_sa *child,
 	bool const is_seclabel_required = (c->spd.this.sec_label.len != 0);
 	chunk_t const *const selected_sec_label =
 		score_ends_seclabel(&e, c, &tsi, &tsr, child->sa.st_logger);
-	if (is_seclabel_required) {
-		if (!selected_sec_label) {
-			/* Security label required, but not found. */
-			return false;
-		}
-	} else {
-		if (selected_sec_label) {
-			/* Security label *not* required, but found. */
-			return false;
-		}
+	if (is_seclabel_required != (selected_sec_label != NULL)) {
+		/*
+		 * Either:
+		 *  - Security label required, but not found.
+		 *    OR
+		 *  - Security label *not* required, but found.
+		 */
+		return false;
 	}
 
 	struct best_score best = score_ends_iprange(initiator_widening, c, &e, &tsi, &tsr);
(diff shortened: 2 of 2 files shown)


cagney (changes requested):
I would change score_ends_seclabel() to return the «chunk_t *» - returning both a bool and the value seems redundant.

(there's a bunch of issues related to memory and efficiency but they aren't specific to this patch)

cagney:
> (there's a bunch of issues related to memory and efficiency but they aren't specific to this patch)

To be clear, approving this change isn't predicated on addressing these.  I'm just noting them for reference:

- adding score_ends_seclabel() means the traffic selectors are scanned twice (the current algorithm is O(N^2)
- fortunately there's a hard limit on the nr of traffic selectors accepted this is bounded; there's a plan to re-work the TS code so something to look at then
- the code parsing the sec_label should look more like 17cc1ea6fddaf028aaa4fcd5d3f7eb6fecdae440 however ...
- instead of cloning (and then leaking) the sec_label in the payload it should take a reference using pbs_in_left_as_shunk()
(this would mean score_ends_seclabel() returns shunk_t*) it gets complicated though, there's code storing the TS in struct state

HughR:
This:
		if (child->sa.st_seen_sec_label.len != 0) {
			free_chunk_content(&child->sa.st_seen_sec_label);
		}

This could be simplified to:

		free_chunk_content(&child->sa.st_seen_sec_label);

HughR:
const chunk_t* selected_sec_label = NULL;
Our style would write this as:
 		const chunk_t *selected_sec_label = NULL;

HughR:
> I would change score_ends_seclabel() to return the «chunk_t *» - returning both a bool and the value seems redundant.

Is it always the case that *selected_sec_label is NULL iff score_ends_seclabel returns false?

cagney:
> Is it always the case that *selected_sec_label is NULL iff score_ends_seclabel returns false?

It should be.

HughR:
There are two versions of score_ends_seclabel.  One always assigns to *selected_sec_label if selected_sec_label is non-NULL.  The other never assigns to it.  This means that the two implementations are not interchangeable.  That seems unnecessarily confusing.

The declaration
   const chunk_t **selected_sec_label,
is quite correct, but misleading.  **selected_sec_label is const BUT the function always writes to *selected_sec_label.
It might help to add a comment on that declaration.  Perhaps:
  const chunk_t **selected_sec_label, /* result parameter */

If Andrew's suggestion works, a few things are improved: the pointer-to-pointer type (that bothers Paul) is eliminated, the confusing result parameter is eliminated, and the code is simplified.

kavinda-forcepoint:
> I would change score_ends_seclabel() to return the «chunk_t *» - returning both a bool and the value seems redundant.

How would we handle the situation where labeled IPsec support is compiled in but not used at runtime? In that case, `score_ends_seclabel()` should return `true` but there is no selected security label.

letoams:
On Wed, 24 Feb 2021, kavinda-forcepoint wrote:

>       I would change score_ends_seclabel() to return the «chunk_t *» - returning both a bool and the value seems redundant.
> 
> How would we handle the situation where labeled IPsec support is compiled in but not used at runtime? In that case,
> score_ends_seclabel() should return true but there is no selected security label.

If the connection has no policy-label, then score_ends_seclabel() should
not be called ? It is not a consideration. And receiving a SEC_LABEL
should return a TS_UNACCEPTABLE?

kavinda-forcepoint:
> If the connection has no policy-label, then score_ends_seclabel() should not be called ? It is not a consideration.

Does that mean I should wrap  each `score_ends_seclabel()` call with a `if (<connection>->spd.this.sec_label.len != 0) { ... }` block?

> And receiving a SEC_LABEL should return a TS_UNACCEPTABLE?

To implement this behavior, should I just return `false` in `v2_parse_ts()` if it encounters a `TS_SECLABEL` when the connection doesn't have a `policy-label`? If yes, this would require me to pass the `struct connection` into `v2_parse_ts()`.

cagney:
> > If the connection has no policy-label, then score_ends_seclabel() should not be called ? It is not a consideration.
> 
> Does that mean I should wrap each `score_ends_seclabel()` call with a `if (<connection>->spd.this.sec_label.len != 0) { ... }` block?

This is getting messy.

> > And receiving a SEC_LABEL should return a TS_UNACCEPTABLE?
> 
> To implement this behavior, should I just return `false` in `v2_parse_ts()` if it encounters a `TS_SECLABEL` when the connection doesn't have a `policy-label`? If yes, this would require me to pass the `struct connection` into `v2_parse_ts()`.

I'd add a bool to structure passed in to v2_parse_ts*() flaging a traffic selector blob with a sec_label.  That way v2_parse_ts() sticks to parsing.

The if() would then look something like:
```
if (spd want's sec_label)
   sec_label = score_ends_seclabel()
   if no-label
    continue
else
  if traffic selector blob contains sec label
    continue
```

letoams:
On Wed, 24 Feb 2021, kavinda-forcepoint wrote:

>       If the connection has no policy-label, then score_ends_seclabel() should not be called ? It is not a consideration.
> 
> Does that mean I should wrap each score_ends_seclabel() call with a if (<connection>->spd.this.sec_label.len != 0) { ... } block?

either wrap it or fold it in at the start of the function? Just make
sure that:

- If the connection has no label, and we got a label (from IKE or
   ACQUIRE), the entire Traffic Selector should fail.

- If the connection has a label, and we did not get a label (from IKE
or ACQUIRE), then entire Traffic Selector should fail.

>       And receiving a SEC_LABEL should return a TS_UNACCEPTABLE?
> 
> To implement this behavior, should I just return false in v2_parse_ts() if it encounters a TS_SECLABEL when the connection doesn't have
> a policy-label? If yes, this would require me to pass the struct connection into v2_parse_ts().

I believe the policy label of the connection residing in struct end is
converted to struct traffic_selectors and passed in? And then compared
against the received TS from struct payload_digest *const ts_pd ?

At least that was the case. Perhaps that is the part you changed, in
which case perhaps just pass both in?

Paul

kavinda-forcepoint:
@cagney and @letoams: I updated the code based on your review comments. Please let me know if the latest changes suffice.

HughR:
Thanks for the recent update to your patch.

The code has several copies of code like this:

	       if (is_seclabel_required) {
			if (!selected_sec_label) {
				/* Security label required, but not found. */
				continue;
			}
		} else {
			if (selected_sec_label) {
				/* Security label *not* required, but found. */
				continue;
			}
		}

1.  One style rule in libreswan is to not  treat a pointer or an int as a boolean (even though that is perfectly correct C).  So the test "select_sec_label" should be "select_sec_label != NULL" instead.
2. This sequence can be simplified
        /* requirement of seclabel must match presense of seclabel */
        if (is_seclabel_required != (selected_sec_label != NULL))
              continue;

cagney (commented):
ok

kavinda-forcepoint:
@HughR: Thank you for the feedback. For the comment, did you mean to say "requirement of seclabel must match *presence* of seclabel", i.e. "presense" is a typo?

HughR:
> @HughR: Thank you for the feedback. For the comment, did you mean to say "requirement of seclabel must match _presence_ of seclabel", i.e. "presense" is a typo?

Yes.  Thanks for catching that.

kavinda-forcepoint:
I have updated the patch after incorporating @HughR's feedback.

kavinda-forcepoint:
I also fixed another bug involving SPD entries as listed above.

letoams:
This has been merged. I still have some questions but we can discuss that separately

cagney:
FYI,

>     * the code parsing the sec_label should look more like [17cc1ea](https://github.com/libreswan/libreswan/commit/17cc1ea6fddaf028aaa4fcd5d3f7eb6fecdae440) however ...

I''ve fixed this.

>     * instead of cloning (and then leaking) the sec_label in the payload it should take a reference using pbs_in_left_as_shunk()
>       (this would mean score_ends_seclabel() returns shunk_t*) it gets complicated though, there's code storing the TS in struct state

I've fixed this.

PR #420, opened 2021-03-02
fix issues in IKEv2 labeled IPsec

kavinda-forcepoint:
### pluto: remove equality check involving SELinux labels
* Any decision involving SELinux labels should always be SELinux policy
  driven.
  ** Therefore, `pluto` should never do equality checks on labels.
* In the case of `within_range()`, there is no guarantee that a label is
  allowed to `polmatch` against itself.
  ** Hence, there should not be a `hunk_eq()` call.

### pluto: fix bug where Initiator only sent a security label in TSi

* Per IKEv2 labeled IPsec proposal (https://tools.ietf.org/html/draft-ietf-ipsecme-labeled-ipsec-04),
  both `TSi` and `TSr` contain a `TS_SECLABEL` payload when security
  labels are in use.
  ** For SELinux, the security label payloads in `TSi` and `TSr` should be
     identical.
* Prior to this change, the Initiator did _not_ send a `TS_SECLABEL`
  payload as part of the `TSr` payload.

### pluto: fix bug where the wrong label was used for SELinux policy checko

* The bug was that `policy-label` (e.g. `ipsec_spd_t`) was used in place
  of the SELinux domain in the `TS_SECLABEL` payload in `TSr`.

### pluto: use `policy-label` value for initial child/IPsec SA pair's label

* A previous commit made the initial child/IPsec SA pair _not_ use a
  label.
* Not using a label, however, caused traffic to bypass the IPsec tunnel
  when using labeled IPsec at just one endpoint, i.e. `policy-label` was
  only specified at one endpoint.
* In order to prevent this "leakage", the initial child/IPsec SA pair
  uses the `policy-label` value for the label, which is usually
  `ipsec_spd_t`.
  ** The behavior implemented by this commit was, in fact, the original
     behavior of the IKEv2 labeled IPsec implementation in Libreswan.
  ** Previously, we had to add the following SELinux rule to the
     `targeted` policy when `ipsec_spd_t` was specified for the initial
     child/IPsec pair:
     `allow ipsec_spd_t self:association { polmatch };`
  ** The rule is no longer needed because:
     *** `pluto`'s code-path no longer invokes a `polmatch` check for the
         initial child/IPsec SA pair via `se_label_match()`.
     *** The kernel creates a child/IPsec SA pair with `ipsec_spd_t`
         without complaint - I am not sure why this is though.

### testing: update expected output for ikev2-labeled-ipsec-03-multi-acquires-enforced

* The output needed to be updated since the test now does what is expected.

### testing: update expected output for ikev2-labeled-ipsec-03-multi-acquires-enforced

* Update `post-mortem.sh` to generate possibly helpful reference policy in case
  there are SELinux AVC (Access Vector Cache) errors.

###  testing: don't remove the SELinux policy after test run in ikev2-labeled-ipsec-03-multi-acquires-enforced

* Removing the SELinux policy _before_ `pluto` is stopped in `post-mortem.sh`
  causes AVC (Access Vector Cache) denials to appear in the audit log.

### testing: update expected output for ikev2-labeled-ipsec-03-multi-acquires-enforced

* SELinux policy is no longer removed at the end of the test.

###  testing: dump SELinux reference policy for missing rules into a separate file


letoams:
It is required for the first conn that does not have an acquire from either side. Otherwise, that conn will always fail with the standard targeted selinux-policy

kavinda-forcepoint:
> It is required for the first conn that does not have an acquire from either side. Otherwise, that conn will always fail with the standard targeted selinux-policy

I looked into this and thought of two possible solutions:

1. Modify the `targeted` SELinux policy to include `allow ipsec_spd_t ipsec_spd_t:association { polmatch }`.
    - This doesn't quite make sense since `ipsec_spd_t` isn't a SELinux domain and therefore unlikely to ever be used for transmitting network traffic.
    - Therefore, adding the above rule to the SELinux policy is more of a hack/workaround than an actual solution.

2. Don't use any labels for the child/IPsec SA pair created during `IKE_AUTH` when there is no `ACQUIRE` involved.
    - This means there is a child/IPsec SA pair without any labels.
    - The child/IPsec SA pair without labels does _not_ get used. That is because every IP datagram when labeled IPsec is enabled has a SELinux domain associated with it; hence, none of the datagrams match the child/IPsec SA pair without any labels.
    - Therefore, the unused unlabled child/IPsec SA pair is harmless.
    - This was the behavior in the original IKEv1 labeled IPsec implementation as well as in the one I initially implemented for IKEv2.

I implemented solution 2 in the latest series of commits above. Let me know what you think.

letoams:
I think there now might be an issue if one end is configured with labels, and the other end is not. The first connection will not come up without label negotation, and the end configured without labeling will actually send traffic over that IPsec SA ?

letoams:
merged. Thanks!

kavinda-forcepoint:
> merged. Thanks!

I noticed the `hunk_eq()` elimination commit didn't get merged. Was that unintentional or on purpose?

letoams:
On Thu, 4 Mar 2021, kavinda-forcepoint wrote:

> I noticed the hunk_eq() elimination commit didn't get merged. Was that unintentional or on purpose?

unintentional. I'll pick it up.

kavinda-forcepoint:
> I think there now might be an issue if one end is configured with labels, and the other end is not. The first connection will not come up without label negotation, and the end configured without labeling will actually send traffic over that IPsec SA ?

This is a good point - thanks for bringing it up. I will investigate and let you know what I find.

letoams:
I pulled in the patch.

Installing a child/IPsec SA is needed to get the XFRM policy to generate the ACQUIRES.
We could special case things and not install anything except the outgoing SPD policy, but that makes things hard. So I prefer to just install it as if it is a normal SA even if it is never actually used.

cagney (commented):
It seems strange to invoke ausearch twice.

My hunch is that it is to dump the "SELinux policy allow/dontaudit rules" making it easier to go back and add them later?  I'd comment this, and dump the output into a separate file?  For instance OUTPUT/${hostname}.audit2allow.log or .rules or ...

kavinda-forcepoint:
> My hunch is that it is to dump the "SELinux policy allow/dontaudit rules" making it easier to go back and add them later?

Yes

> I'd comment this, and dump the output into a separate file? For instance OUTPUT/${hostname}.audit2allow.log or .rules or ...

Good idea.

cagney (approved):

PR #431, opened 2021-04-11
OpenBSD Fixes

RaviTejaCMS:
Make kvm-install should now automatically build OpenBSD domains


letoams:
There were some weird merge issues with this.
I cherry-picked one of the commits from this list and did some other manual fixing.
The last thing left is the "east_variant" thing, that seems to be missing in the tree yet is being modified by this PR?

cagney:
> The last thing left is the "east_variant" thing, that seems to be missing in the tree yet is being modified by this PR?

It isn't needed.

It was so that obsdeinit.sh ran early.  OpenBSD tests can just use numbered test scripts instead.

letoams:
On Tue, 13 Apr 2021, cagney wrote:

>       The last thing left is the "east_variant" thing, that seems to be missing
>       in the tree yet is being modified by this PR?
> 
> It isn't needed.
> 
> It was so that obsdeinit.sh ran early. OpenBSD tests can just use numbered test
> scripts instead.

ahh, I see. so you removed it :)

Well, it seems with current git main, I can run kvm-demolish and
kvm-install and get working openbsd domains and pass openbsd tests.

cagney:
> > t was so that obsdeinit.sh ran early. OpenBSD tests can just use numbered test scripts instead.
> ahh, I see. so you removed it :)

I don't think it was me.  I just noticed half the change was missing, and figured it wasn't needed.

I've filed two bugs.

letoams:
the test had both an eastinit.sh and numbered scripts, so it was doing an init phase on east twice. I did fix that part. It might have been a merge anomaly ?

PR #437, opened 2021-04-19
Bitstomask6 wrong for slash zero

mcr:
words 1,2,3 might be zero by chance, but no guarantee.
obvious copy and paste typo.  Clearly few people use ::/0 policies.

### DIFF
--- a/linux/net/ipsec/goodmask.c
+++ b/linux/net/ipsec/goodmask.c
@@ -156,9 +156,9 @@ int n;
 		result.s6_addr32[3] = htonl(~((1UL << (128 - n)) - 1));
 	} else {
 		result.s6_addr32[0] = 0;
-		result.s6_addr32[0] = 0;
-		result.s6_addr32[0] = 0;
-		result.s6_addr32[0] = 0;
+		result.s6_addr32[1] = 0;
+		result.s6_addr32[2] = 0;
+		result.s6_addr32[3] = 0;
 	}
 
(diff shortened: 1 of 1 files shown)


cagney:
FYI, KLIPS was removed from libreswan 4.0.

mcr:
Well, it's still in "master".  Have you moved to "main" then? And anyway, the patch applies to wherever goodmask.c has moved to.

letoams:
There hasn’t been a master branch on the GitHub page for many months.

Thanks for sharing the bug, luckily it doesn’t apply to us.

Paul

HughR:
Nice catch.  That's been there since you committed it in 2007!
The pattern is really odd.  Each subsequent case handles larger and larger n, but the last one handles 0 (and all other unhandled values, but 0 is probably the only legit one).
How about unconditionally zeroing all the members, before the IFs, then eliminate all the explicit assignments of 0, and then the empty ELSE.  Makes the code shorter and clearer.

Something like this might work (untested):
	for (int i = 0; i < 4; i++) {
		result.s6_addr32[i] =
                        n < 0 ? 0UL :
                        n < 32 ? htonl(~0UL << n) :
 	 	        ~0UL;
		n -= 32;
	}

letoams:
This PR is against old deleted code

mcr:
Libreswan swill has goodmask.3.xml lurking.

PR #438, opened 2021-04-23
Fix tests on all systems

kekePower:
Hi.

I got a compilation error on arm and aarch64 and asked around for help.

After applying this patch, libreswan built fine on Mageia Cauldron on all systems.

Thanks to jeeebz on irc mageia-dev for the patch


cagney (approved):
oops!

bleve:
Thx, committed in.

PR #444, opened 2021-05-12
pluto: unshare policy labels in connection `end`s

kavinda-forcepoint:
* Prior to this fix, `pluto` would trigger a double free error when
  cleaning up `connection`s.


paulwouters:
So I think all parts of this PR have been merged in? I'm not sure why github does not see that?

kavinda-forcepoint:
> So I think all parts of this PR have been merged in? I'm not sure why github does not see that?

You are right - I see the PR merged to mainline in its commit history. Not sure why that does not get reflected here. I will go ahead and close this PR then.

PR #452, opened 2021-05-22
Add missing include statements.

yurivict:


cagney:
Is this a bug in the freebsd headers?

    for u_short that netinet/udp.h needs on FreeBSD

everything's merged

yurivict:
> Is this a bug in the freebsd headers?
> for u_short that netinet/udp.h needs on FreeBSD

Not necessarily. Usually function manpage says what headers to include and they sometimes come in pairs or triplets.
Individual headers don't have to be self-sufficient.

yurivict:
Thanks for merging.

Issue #453, opened 2021-05-27
IKE_AUTH redirect response contains CHILD SA


cagney:
| **parse ISAKMP Message:
|    initiator SPI: d7 43 4f 0a  b3 b0 a6 15
|    responder SPI: fa bd b8 38  df bb e0 03
|    next payload type: ISAKMP_NEXT_v2SK (0x2e)
|    ISAKMP version: IKEv2 version 2.0 (rfc4306/rfc5996) (0x20)
|    exchange type: ISAKMP_v2_IKE_AUTH (0x23)
|    flags: ISAKMP_FLAG_v2_MSG_RESPONSE (0x20)
|    Message ID: 1 (00 00 00 01)
|    length: 239 (00 00 00 ef)
|  processing version=2.0 packet with exchange type=ISAKMP_v2_IKE_AUTH (35)
| I am the IKE SA Original Initiator receiving an IKEv2 IKE_AUTH response 
| State DB: found IKEv2 state #1 in PARENT_I2 (find_v2_ike_sa)
| #1 is idle
| #1 idle
| unpacking clear payload
| Now let's proceed with payload (ISAKMP_NEXT_v2SK)
| ***parse IKEv2 Encryption Payload:
|    next payload type: ISAKMP_NEXT_v2N (0x29)
|    flags: none (0x0)
|    length: 211 (00 d3)
| processing payload: ISAKMP_NEXT_v2SK (len=207)
| #1 in state PARENT_I2: sent IKE_AUTH request
| stripping 1 octets as pad
| #1 ikev2 ISAKMP_v2_IKE_AUTH decrypt success
| Now let's proceed with payload (ISAKMP_NEXT_v2N)
| **parse IKEv2 Notify Payload:
|    next payload type: ISAKMP_NEXT_v2IDr (0x24)
|    flags: none (0x0)
|    length: 14 (00 0e)
|    Protocol ID: IKEv2_SEC_PROTO_NONE (0x0)
|    SPI size: 0 (00)
|    Notify Message Type: v2N_REDIRECT (0x4017)
| processing payload: ISAKMP_NEXT_v2N (len=6)
| Now let's proceed with payload (ISAKMP_NEXT_v2IDr)
| **parse IKEv2 Identification - Responder - Payload:
|    next payload type: ISAKMP_NEXT_v2AUTH (0x27)
|    flags: none (0x0)
|    length: 12 (00 0c)
|    ID type: ID_FQDN (0x2)
|    reserved: 00 00 00
[...]

PR #467, opened 2021-06-25
ikev2 redirect: make peer redirecting (others) in IKE_AUTH childless

vukasink:
This solves #453 as well. Also peers that are redirecting others don't have a dead kernel policy anymore, see testing commit.


paulwouters:
merged. thanks!

Issue #474, opened 2021-07-19
Compile error with OpenSSL 3.0, due to conflicting macros with NSS


ueno:
When compiled with OpenSSL 3.0, I get:
```console
In file included from /usr/include/nss3/pk11pub.h:11,
                 from /builddir/build/BUILD/libreswan-4.2/programs/pluto/state.h:46,
                 from /builddir/build/BUILD/libreswan-4.2/programs/pluto/ikev2_ipseckey.c:34:
/usr/include/nss3/certt.h:562: error: "KU_DIGITAL_SIGNATURE" redefined [-Werror]
  562 | #define KU_DIGITAL_SIGNATURE (0x80) /* bit 0 */
      | 
In file included from /usr/include/openssl/ocsp.h:63,
                 from /usr/include/openssl/x509.h:626,
                 from /usr/include/openssl/ssl.h:31,
                 from /usr/include/ldns/dane.h:30,
                 from /usr/include/ldns/ldns.h:98,
                 from /builddir/build/BUILD/libreswan-4.2/programs/pluto/ikev2_ipseckey.c:26:
/usr/include/openssl/x509v3.h:662: note: this is the location of the previous definition
  662 | # define KU_DIGITAL_SIGNATURE    0x0080
      | 
In file included from /usr/include/nss3/pk11pub.h:11,
                 from /builddir/build/BUILD/libreswan-4.2/programs/pluto/state.h:46,
                 from /builddir/build/BUILD/libreswan-4.2/programs/pluto/ikev2_ipseckey.c:34:
/usr/include/nss3/certt.h:563: error: "KU_NON_REPUDIATION" redefined [-Werror]
  563 | #define KU_NON_REPUDIATION (0x40)   /* bit 1 */
      | 
...
```
It seems that the same macros are defined in both NSS and OpenSSL headers indirectly included through `<ldns/ldns.h>`. I guess an easy fix would be to undef those, but it might make sense to isolate the uses of `<ldns/ldns.h>` into a separate module so it doesn't interfere once and for all.

paulwouters:
I merged these in. The two small commits were fine, the big commit needed some work due to @cagney making some recent changes, so I added a Co-authored-by: to take the blame for any issues introduced :)
I ran the dnsoe tests, so it looks good to me.

ueno:
Thank you for the review and adjustment!

PR #475, opened 2021-07-19
ikev2 ipseckey: split out DNS query code into a separate file

ueno:
When compiled with DNSSEC support, `<ldns/ldns.h>` indirectly pulls in OpenSSL headers where symbols conflicting with the ones defined in NSS headers, such as `KU_*` macros.  This moves out the low-level DNS lookup code using LDNS to a dedicated file, so there will be no interference.

Fixes: #474


paulwouters:
Merged. Thanks!

PR #477, opened 2021-07-22
ipsec barf: fix shell test expression

ueno:
Spotted by shellcheck:
```
  /usr/libexec/ipsec/barf:55:5: error[SC1073]: Couldn't parse this test expression. Fix to allow more checks.
  #   53|       for f
  #   54|       do
  #   55|-> 	if [ -s ${LOGS}/${f} -a \
  #   56|   	    -f ${LOGS}/${f} -a \
  #   57|   	    grep -E -q "${s}" ${LOGS}/${f} 2>/dev/null ]
```
Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/programs/barf/barf.in
+++ b/programs/barf/barf.in
@@ -53,8 +53,8 @@ findlog() {	# findlog string fallbackstring possiblefile ...
     for f
     do
 	if [ -s ${LOGS}/${f} -a \
-	    -f ${LOGS}/${f} -a \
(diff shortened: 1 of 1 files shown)


bleve:
Sorry but that is completely valid syntax and it is a shellcheck bug it doesn't understand that. It is tested to work in all shells and there is no syntactic errors.

bleve:
Sorry. I was wrong - that actually fails. Will apply. Thx.

ueno:
Thanks for the review and merging it!

PR #478, opened 2021-07-23
pluto: fix bug in labeled IPsec and peer wildcard IP interaction

kavinda-forcepoint:
This fix implements the following desired behavior:

- A connection configured with a peer wildcard IP (`%any`) and labeled
  IPsec (`policy-label`) is a template (`CK_TEMPLATE`) for both of these
  reasons. Call this template T1.
- When that T1 gets instantiated with just an actual peer IP, the new
  `struct connection` needs to also be a template due to its label being
  the same as `policy-label`. Call this new template T2.
- T2 is later instantiated with a specific label used for network
  transmission (i.e. NOT `policy-label`) to get an instance
  (`CK_INSTANCE`).


kavinda-forcepoint:
Closing PR since the fix was merged in https://github.com/libreswan/libreswan/commit/cdbf203f8411f663f083d8609b2d4d9d4ac70d24

PR #483, opened 2021-07-29
pluto: fix memory leak + minor refinements in within_range()

kavinda-forcepoint:
* Memory allocated by `getcon()` could leak if function returned early.
* Replace `strerror()` with `strerror_r()` for thread safety.
* Update debugging output.


paulwouters:
merged

PR #486, opened 2021-08-07
pluto: unshare connection::v2_ike_auth_child_proposals

kavinda-forcepoint:
* Prior to this fix, multiple `connection`s'
  `v2_ike_auth_child_proposals` could point to the same `struct`.
* Therefore, connection cleanup could trigger a double free error as
  multiple `connection`s repeatedly free the same `struct`.


cagney:
Is there a test / log?  I'm puzzled as to why a template connection has .v2_ike_auth_child_proposals (a connection instance thing) non-NULL.

cagney:
Several thoughts:

- it could easily be from all the Child SA changes, however do I wonder if cdbf203f8411f663f083d8609b2d4d9d4ac70d24 is leading to .v2_ike_auth_child_proposals being set on the intermediate template connection instead of the final connection instance?  This would be a bigger problem but one a passert(c is an instance) in get_v2_ike_auth_child_proposals() would easily detect
- the test ikev2-labeled-ipsec-03-multi-acquires creates a childless SA, turning it into ikev2-labeled-ipsec-03-multi-acquires-ike-{ondemand,childless} would likely help here (the former hopefully triggers the bug); I can do that?

As an aside, I suspect it is now possible for v2_ike_auth_child_proposals to be generated during connection load and the stored in struct config.  It depends on policy, but as best I can tell those policy bits don't change.

kavinda-forcepoint:
> Is there a test / log? I'm puzzled as to why a template connection has .v2_ike_auth_child_proposals (a connection instance thing) non-NULL.

I neither have a test nor a log. I found this when doing IKE rekey testing. The segfault due to the double free happened after several rounds of re-keying.

kavinda-forcepoint:
> Several thoughts:
> 
>     * it could easily be from all the Child SA changes, however do I wonder if [cdbf203](https://github.com/libreswan/libreswan/commit/cdbf203f8411f663f083d8609b2d4d9d4ac70d24) is leading to .v2_ike_auth_child_proposals being set on the intermediate template connection instead of the final connection instance?  This would be a bigger problem but one a passert(c is an instance) in get_v2_ike_auth_child_proposals() would easily detect

I will add this `passert()` and let you know what happens.

>     * the test ikev2-labeled-ipsec-03-multi-acquires creates a childless SA, turning it into ikev2-labeled-ipsec-03-multi-acquires-ike-{ondemand,childless} would likely help here (the former hopefully triggers the bug); I can do that?

Does that test do IKE rekeying? As I mentioned above, the bug manifested after several rounds of IKE re-keying.

kavinda-forcepoint:
@cagney: I added the `passert()` you suggested and it triggered due to `c->kind == CK_TEMPLATE`. Here's the backtrace:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f2cc8935db5 in __GI_abort () at abort.c:79
#2  0x0000562cdfd83a49 in llog_passert (logger=0x562ce00476e0 <failsafe_logger>, where=where@entry=0x562ce003d770 <here.18753>, fmt=fmt@entry=0x562cdfde62f6 "%s")                                                
    at lib/libswan/passert.c:34
#3  0x0000562cdfd55f57 in get_v2_ike_auth_child_proposals (c=c@entry=0x562ce0657748, why=why@entry=0x562cdfdb6268 "IKE SA initiator emitting ESP/AH proposals", logger=<optimized out>)                           
    at programs/pluto/ikev2_spdb_struct.c:2199                                                                                              
#4  0x0000562cdfcd4246 in ikev2_in_IKE_SA_INIT_R_or_IKE_INTERMEDIATE_R_out_IKE_AUTH_I_signature_continue (ike=0x562ce06990b8, md=<optimized out>, auth_sig=<optimized out>)                                       
    at programs/pluto/ikev2_ike_auth.c:457                                                                                                  
#5  0x0000562cdfd23861 in handle_helper_answer (st=0x562ce06990b8, md=0x562ce0692328, arg=<optimized out>)                                                                                                        
    at programs/pluto/server_pool.c:445                                                                                                     
#6  0x0000562cdfd21466 in resume_handler (fd=<optimized out>, events=<optimized out>, arg=0x7f2cbc0135d8) at programs/pluto/server.c:654    
#7  0x00007f2cc935ed85 in event_process_active_single_queue (base=base@entry=0x562ce05a6a18, activeq=0x562ce0647c88, max_to_process=max_to_process@entry=2147483647, endtime=endtime@entry=0x0) at event.c:1646   
#8  0x00007f2cc935f787 in event_process_active (base=0x562ce05a6a18) at event.c:1738
#9  event_base_loop (base=0x562ce05a6a18, flags=flags@entry=0) at event.c:1961
#10 0x0000562cdfd22221 in run_server (conffile=0x562ce0590508 "/etc/ipsec.conf", logger=0x7fffd7037e70) at /usr/src/debug/libreswan-4.4-0.rc785_g22d66d3970_labeled_ipsec.x86_64/programs/pluto/server.c:1072     
#11 0x0000562cdfccf910 in main (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/libreswan-4.4-0.rc785_g22d66d3970_labeled_ipsec.x86_64/programs/pluto/plutomain.c:1783
```

Let me know if you need any more information.

cagney:
Something like this might trigger the double free:

- establish IKE SA which, assuming no fragmentation, is two outgoing and two incoming messages
- tell pluto to drop the first attempt at sending message 3: ipsec whack --impair drop-outgoing:3
- trigger message three, a rekey: ipsec whack --rekey-ike ... --async
- watch pluto.log waiting for indication that the message was dropped; the clock is ticking
- trigger acquire with single ping; should be put onto the pending queue
- watch pluto.log waiting for a re-transmit to kick start the stalled rekey and then establish the child
- a shutdown should trigger the double free

If things happen too quick, adding a second <<ipsec whack --impair drop-outgoing:3>> command will cause the first retransmit of message 3 to be dropped.

cagney:
> Something like this might trigger the double free:
[snip]
it wasn't sufficient, but did turn up another problem - acquire during ike rekey was lost; that is fixed

bleve (changes requested):
Makefile part is not ok. It is clearly wrong.

lib/libswan/DBG_dump.c
```diff
@@ -64,4 +66,24 @@ void DBG_dump(const char *label, const void *p, size_t len)
...
+	if (label == NULL) {
+		return;
+	}
+
+	const char *ch = p;
+	for (size_t i = 0; i < init_len; i++) {
+		if (*ch == '\0') {
+			break;
+		}
+		if ((*ch < ' ') || (*ch > '~')) {
```
paulwouters, line 79: There seems to be some assumption here about space or - that is new and not AFAIK in the current code for security labels ?

lib/libswan/DBG_dump.c
```diff
@@ -64,4 +66,24 @@ void DBG_dump(const char *label, const void *p, size_t len)
...
+			return;
+		}
+		++ch;
+	}
+
+	char* buf = alloc_bytes(init_len + 1, "dump buf");
```
paulwouters, line 85: Maybe create a chunk_t with ptr and len, then use DBG_dump_hunk()  so that there is no need for alloc/free ?

packaging/rhel/8/libreswan.spec
```diff
@@ -27,7 +27,7 @@
     USE_LIBCURL=true \\\
     USE_LINUX_AUDIT=true \\\
     USE_NM=true \\\
-    USE_NSS_KDF=true \\\
+    USE_NSS_KDF=false \\\
```
paulwouters, line 30: these seem accidental, like the kvm-targets.mk and makefile changes. But why did you need to set this to false? As rhel8/centos8 should already contain new enough nss libraries to leave this set to true

kavinda-forcepoint:
Ugh, sorry @bleve and @paulwouters, this was an accidental commit that got pushed. I will close this PR for now since @cagney has been working on an alternate solution. If we need my original commit again, I will create a new PR.

packaging/rhel/8/libreswan.spec
kavinda-forcepoint: I will try reverting to see if it works. I can't recall, but this might be an artifact from when I was working on RHEL 8.2.

PR #506, opened 2021-09-02
pluto: fix build warnings against musl c-library

fabled:
Use POSIX defined locations for poll.h and sys/wait.h.
Cast cmsg_len to size_t as the struct member may be an int.


paulwouters:
merged, thanks

Issue #588, opened 2021-12-28
update documentation for `ikev2` config var


dkg:
`configs/d.ipsec.conf/ikev2.xml` says:

```
  <varlistentry>
  <term><emphasis remap='B'>ikev2</emphasis></term>
  <listitem>
<para>Whether to use IKEv1 (RFC 4301) or IKEv2 (RFC 7296) settings to be used.
Currently the accepted values are <emphasis remap='B'>no</emphasis>(the default),
signifying only IKEv1 is accepted, or <emphasis remap='B'>yes</emphasis>,
signifying only IKEv2 is accepted. Previous versions allowed the keywords
<emphasis remap='B'>propose</emphasis> or <emphasis remap='B'>permit</emphasis>
that would allow either IKEv1 or IKEv2, but this is no longer supported. The
permit option is interpreted as no and the propose option is interpreted as
yes. Older versions also supported keyword
<emphasis remap='B'>insist</emphasis> which is now interpreted as yes.
</para>
  </listitem>
  </varlistentry>
```

Two things need cleaning up:

- The first sentence is confusing English.
- it claims that the default is "no", but `ipsecconf_default_values()` (in `lib/libipsecconf/`) sets:

```
	d->ike_version = IKEv2;
```

which i think is the variable that it controls.

PR #589, opened 2021-12-28
ipsec.conf docs: clean up documentation about "ikev2" config var.

dkg:
Please verify that this update is correct!

Closes: #588


dkg:
I think this was merged (with a minor modification) by Tuomo, in 3783ea5a843334d3ef5b9262ff2be467eb868ce5

PR #590, opened 2021-12-28
More orthographic fixes

dkg:
The first three commits are just typo/spelling fixes with no substantive impact.

the last two commits are typo/spelling fixes that actually correct some python code in testing/


dkg:
I think all of these have been merged in one form or another by Tuomo Soini, so this can probably be closed.

PR #606, opened 2022-01-08
Correct warnings when USE_IKEv1 is not defined

dkg:
The logic was inverted here.  If USE_IKEv1 is not defined, then we
should warn/log only when pluto_ikev1_pol is *not* GLOBAL_IKEv1_DROP.

### DIFF
--- a/programs/pluto/plutomain.c
+++ b/programs/pluto/plutomain.c
@@ -1198,7 +1198,7 @@ int main(int argc, char **argv)
 			pluto_ddos_mode = cfg->setup.options[KBF_DDOS_MODE];
 			pluto_ikev1_pol = cfg->setup.options[KBF_GLOBAL_IKEv1];
 #ifndef USE_IKEv1
-			if (pluto_ikev1_pol != GLOBAL_IKEv1_ACCEPT) {
+			if (pluto_ikev1_pol != GLOBAL_IKEv1_DROP) {
 				llog(RC_LOG, logger, "ignoring ikev1-policy= as IKEv1 support is not compiled in. Incoming IKEv1 packets are silently dropped");
 				pluto_ikev1_pol = GLOBAL_IKEv1_DROP;
 			}


paulwouters:
I think you misunderstood the code. There is an if NOT def there.

Basically, this check is to ensure that if no IKEv1 support has been compiled in, that the policy for IKE MUST NOT be set to ikev1-policy=accept. If someone does this, the policy is overwritten to ikev1-policy=drop because we have no code compiled to handle such packets

dkg:
i think i do understand the code, but it's always possible that i'm confused (i'm good at being confused! ☺)


I had written my patch assuming that `drop` was the only acceptable setting when `USE_IKEv1` was not defined.  That is, that `reject` would be unacceptable because we don't want to even have the code required to parse the rejection destination out of the incoming IKEv1 packet.  But it sounds like you're saying that `reject` is also acceptable when `USE_IKEv1` is not defined.

If that's the case, let me try restating the stanza in question:

in the event that `USE_IKEv1` is unset, we *must* not have `ikev1-policy=accept`.  We would be OK with either `drop` or `reject`.  is that right?   but if the user tries to make it `accept` then we should revert to `drop` and warn. if that's the case, then the test is still wrong: it tests:

    if (pluto_ikev1_pol != GLOBAL_IKEv1_ACCEPT) {

But it should test:

    if (pluto_ikev1_pol == GLOBAL_IKEv1_ACCEPT) {

If i'm confused, please help unconfuse me :)

paulwouters:
You were right. Thanks for persisting :)

dkg:
no problem!  looks like aa122b1387966f7c70590e4727795fdc65663237 changed the log line to insert a `PAUL:` prefix.    i don't see any other such prefix in a log line in the codebase.  is that just a debugging thing that slipped through?

letoams:
On Sat, 8 Jan 2022, dkg wrote:

> no problem! looks like aa122b1 changed the log line to insert a PAUL:

Fixed :)

> that just a debugging thing that slipped through?

Yes :)

PR #611, opened 2022-01-11
pluto: do not include NSS headers in ikev2_ipseckey_dnsr.c

ueno:
This file has an inherent dependency on OpenSSL 3.0 headers through
<ldns/ldns.h>, which causes macro conflicts with NSS.

Related to #474.

Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/programs/pluto/ikev2_ipseckey.h
+++ b/programs/pluto/ikev2_ipseckey.h
@@ -1,4 +1,5 @@
 #include "state.h"
+#include "ikev2_ipseckey_dnsr.h" /* for dns_status */
 
 #ifndef _IKEV2_IPSECKEY_H
 #define _IKEV2_IPSECKEY_H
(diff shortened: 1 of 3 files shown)


ueno:
cf. https://gitlab.com/redhat/centos-stream/rpms/libreswan/-/merge_requests/6

cagney (changes requested):
How's this fixing a build problem?

We've now got ikev2_ike_auth.c, via ikev2_ipseckey.h including the implementation specific ikev2_ipseckey_dnsrc.h!

Can I suggest two things:

- clean up ikev2_ipseckey.h:
-- ifdef IKEV2_IPSECKEY_H
-- (C) notice
-- credits, replace state.h with struct ike_sa; et.al.
- review the includes for ikev2_ipsec*.c - system headers, include/ headers, local headers
I'm pretty sure that at least the first change, the problem will go away

PR #613, opened 2022-01-13
Fix CVE-2022-23094 patch for 4.3 and 4.2

dkg:
libreswan 4.2 and 4.3 only have log_diag, not llog_diag.

Note that this change breaks the OpenPGP signature over the CVE
announcement, so it probably needs to be re-signed.


bleve:
Thix, applied with new signature.

Also fixed the actual patch
https://libreswan.org/security/CVE-2022-23094/CVE-2022-23094-libreswan-4.2-4.3.patch

PR #617, opened 2022-01-18
make ikev1-policy default to "drop"

dkg:
IKEv2 has been available for 16 years (RFC 4306 was published December
2005).  At some point, we should be discouraging IKEv1 adoption.

To the extent that a user needs IKEv1, they can manually add
ikev1-policy=accept to /etc/ipsec.conf.

---

This is just a proposal, happy to hear any feedback on it.

At some point, the long term goal will probably be to change `USE_IKEv1` to be undefined by default, but a reasonable first step is to require admins to explicitly turn it on.  that's what this change does.

This kind of deprecation work is a belt-and-suspenders approach to minimize the potential scale of any implementation vulnerabilities like CVE-2022-23094.

### DIFF
--- a/configs/d.ipsec.conf/ikev1-policy.xml
+++ b/configs/d.ipsec.conf/ikev1-policy.xml
@@ -3,9 +3,10 @@
   <listitem>
 <para>
 What to do with received IKEv1 packets. Valid options are
-<emphasis remap='B'>accept</emphasis> (default), <emphasis remap='B'>reject</emphasis> which
-will reply with an error, and <emphasis remap='B'>drop</emphasis> which will silently drop
-any received IKEv1 packet. If this option is set to drop or reject, an attempt to load an
+<emphasis remap='B'>drop</emphasis> (default) which will silently drop
+any received IKEv1 packet, <emphasis remap='B'>accept</emphasis>, and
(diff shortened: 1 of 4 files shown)


cagney:
Using this as a backstop depending on how #1087 is resolved.

cagney:
Yea, every since IKEv1 test needs enable

cagney:
Should `ikev1-policy=accept` instead be dropped.  And only accept when there's an IKEv1 connection.
This is like how groundhog works.

cagney:
Can I configure things so that:
- I can load and allow IKEv1 connections
- drop, rather than reject, other IKEv1 connections?

letoams:
On Wed, 14 Jun 2023, cagney wrote:

> Can I configure things so that:
>
>  *  I can load and allow IKEv1 connections
>  *  drop, rather than reject, other IKEv1 connections?

No, because the first packet of an IKEv1 exchange won't tell you
with certainty you are matching the "right" connection.

Also, the idea is to push people to move to IKEv2.

We changed the default to IKEv2, forcing people to ikev2=no to do
IKEv1. Now we are changing the global config, so they have to
globally re-enable it again or move to ikev2.

if it automatically works for their own ikev1 conns, they will
never move. Then we might as well not have an option at all.

PR #641, opened 2022-02-15
Fix nsspassword location in tests

The-Mule:
NSS password needs to be located in ipsecdir instead
of nssdir. For the testuite it does not matter but
if a testsuite is used with a nssdir that differs from
ipsecdir it causes issues.


bleve:
Thank you. Looks like a correct fix, commited in with 8e5e2bffb5ee0aa8948ff5c241e8eb5e473f83aa

PR #727, opened 2022-05-25
chore: Included githubactions in the dependabot config

nathannaveen:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+      interval: "weekly"


DocEmmetBrown (approved):

cagney:
Given an application built using something like `gradle`, where the dependencies can be hardwired so constantly need to be updated, I can see this as very useful.

Here, however, don't think that applies.  Libreswan isn't built against specific libraries (that's a problem for the downstream packager) so I can see this file effectively being a NOP.

Perhaps there are other uses though:

- with testing
It might be useful to know that there's been a new OS release (for instance f41) so it is time to update that.  But even that has an exception - the debian release should be trailing edge - it only gets updated when it falls off LTS
- with bugs and code
Can it indicate that broken libraries we work-around have been EOLed so we can remove the hacks.

cagney:
Yea,

As I suspected, this causes github to auto-modify the libreswan git repository - for all the documentation nothing spells this out clearly.  The only saving grace is that it creates a branch.  Mainline (which is being pushed from elsewhere) is not affected.

PR #953, opened 2022-12-20
fix debian packaging missing systemd unit ipsec.service

wmasilva:
when "make deb", the resulting package does not have the unit file for systemd, this commit solves the issue.

dpkg -c libreswan_4.9-1_amd64.deb  | grep service
-rw-r--r-- root/root      1531 2022-10-13 01:00 ./lib/systemd/system/ipsec.service

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -27,6 +27,7 @@ override_dh_auto_build:
 		LIBEXECDIR=/usr/libexec/ipsec \
 		MANDIR=/usr/share/man \
 		USE_LDAP=true \
+		INITSYSTEM=systemd \
 		$(ENABLE_LIBCAP_NG) \
 		$(ENABLE_SELINUX)
 
@@ -39,6 +40,7 @@ override_dh_auto_install-arch:
 		LIBEXECDIR=/usr/libexec/ipsec \
 		MANDIR=/usr/share/man \
 		USE_LDAP=true \
+		INITSYSTEM=systemd \
 		$(ENABLE_LIBCAP_NG) \
 		$(ENABLE_SELINUX) \
 		DESTDIR=$(CURDIR)/debian/libreswan


bleve:
Why do you think this is needed? Debian package doesn't have this and works correctly. Could this be you were missing some build dependency on your system?

https://salsa.debian.org/debian/libreswan/-/blob/debian/unstable/debian/rules

wmasilva:
i notice the missing ipsec.service in version 4.8, but in previous version it was ok. 
The link for debian rules you post, is for version 4.7, correct?  

I've check the build, i found the cause,  the variable INITSYSTEM is initialized to "sysvinit" and it should be "systemd".
I'm building the package in a chroot environment and the path "/run/systemd" is not present, so it defaults to "sysvinit".

https://github.com/libreswan/libreswan/blob/1b045f874c596c8796db6ae0545cb0a2626b62e8/mk/defaults/linux.mk#L88

I create the path and it builds ok without changing debian rules.

Probably is a good idea to leave the rules with INITSYSTEM="systemd" so we can build the package with systemd in any environment, for centos/fedora the packaging specs it sets "fedora/libreswan.spec:    INITSYSTEM=systemd "

cagney:
Does `/var/run/systemd` exist?  See fe929e77118 and 8a8623e287d

Anyway, yes debian should force INITSYSTEM, and we might as well change the default from `sysvinit` to `systemd`.  The two big distros have been assimilated.

wmasilva:
yes, /var/run/systemd is present (i created it before), in debian buster /var/run/  is a symlink to /run

[16:10:55][chroot_buster][/]# ls -l /var/run
lrwxrwxrwx 1 root root 4 Aug 23  2019 /var/run -> /run

paulwouters:
merged, thanks!

PR #967, opened 2023-01-02
documentation: Fix typo

qyanu:
Hi,

i'm pretty sure, albeit not 100%, that "send" should be in the "sent" form here.

best regards,
Max

### DIFF
--- a/configs/d.ipsec.conf/nat-keepalive.xml
+++ b/configs/d.ipsec.conf/nat-keepalive.xml
@@ -2,7 +2,7 @@
   <term><emphasis remap='B'>nat-keepalive</emphasis></term>
   <listitem>
 <para>whether to send any NAT-T keep-alives. These one byte packets
-are send to prevent the NAT router from closing its port when there is
+are sent to prevent the NAT router from closing its port when there is
 not enough traffic on the IPsec connection. Acceptable values are:
 <emphasis remap='B'>yes</emphasis> (the default) and
 <emphasis remap='B'>no</emphasis>.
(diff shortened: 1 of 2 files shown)


letoams:
On Mon, 2 Jan 2023, Max-Julian Pogner wrote:

> i'm pretty sure, albeit not 100%, that "send" should be in the "sent" form here.

I don't think so?

The first sentence is about packets that will go out in the future, so "will
be send" seems right to me. Similar for the second sentence, which has
future tense too, eg "which are [to be] send" ?

But I'm not an English native speaker or a ChatGPT bot, so perhaps
someone who has English as first language could confirm this :)

Paul

qyanu:
Hi,

I distinctly remember learning it in school the other way round. But then, teachers err.

Give me a week, maybe i find a generally trusted source[1]. If not, i will silently close this issue and merge-request.

best regards!
Max

[1] I would hate not finding such a source, as then i would in perpetuity occupy my mind with this conundrum.

cagney:
You both are kind of correct.
The sentence uses past tense. But the sentence before uses future tense,
which is probably what should be used through out.
But Strine, and not English is my first language 😄

qyanu:
@letoams
I would say, that this source should be authoritative enough
https://www.bbc.co.uk/worldservice/learningenglish/youmeus/learnit/learnitv65.shtml

They give an example sentence in passive voice: "Three hundred people are employed by the construction company in Station Road."

The similarities with the sentence at hand are striking, and they use the verb "employ" in the past tense form also.

cagney:
_nat-keepalive
whether to send any NAT-T keep-alives. These one byte packets are send to prevent the NAT router from closing its port when there is not enough traffic on the IPsec connection. Acceptable values are:yes  (the default) and no_

what about:

_nat-keepalive
whether to send any NAT-T keep-alives. These one byte packets prevent a NAT router from closing its port when there is not enough traffic on the IPsec connection. Acceptable values are:yes  (the default) and no_

cagney:
I changed it per above.

Issue #747, opened 2022-05-30
openbsd can't build documentation


cagney:
```
mv ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_pluto.8.tmp.tmp ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_pluto.8.tmp
: ignoring seemingly bogus xmlto exit status
xmlto man ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_pluto.8.tmp -o ../../OBJ.openbsd.amd64.openbsd/programs/pluto || true
Writing ipsec_pluto.8 for refentry(pluto8)
Writing ipsec whack.8 for refname
Writing pluto.8 for refname
test -z ""  -a -r ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_pluto.8  -a -r ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_whack.8  -a -r ../../OBJ.openbsd.amd64.openbsd/programs/pluto/pluto.8
gmake[3]: *** [../../mk/manpages.mk:82: ../../OBJ.openbsd.amd64.openbsd/programs/pluto/ipsec_pluto.8.man] Error 1
```
my first guess is xmlto, but who knows

cagney:
Note this line:
```
Writing ipsec whack.8 for refname
```
On Linux the output file name has an underscore as in `ipsec_whack.8`

cagney:
090d4b5b179cfd6704cd4c5cbb3a04e92781249d

Issue #685, opened 2022-04-08
https://libreswan.org/man/ipsec_pluto.8.html says to see the ipsec_pluto.8.html man page


cagney:
https://libreswan.org/man/ipsec_pluto.8.html

jlearman:
Seriously!  Please fix this.  I can't find any documentation for the `whack` command that corresponds to the released code.  6 months?

cagney:
A solution to #848 would let us fix this.  Patches welcome.
In the mean time see `man ipsec whack`.

letoams:
On Fri, 28 Oct 2022, Jeff Learman wrote:

> Seriously! Please fix this. I can't find any documentation for the whack command that corresponds to the released code. 6 months?

There is always "ipsec whack --help"

Note that whack is supposed to be an internal API, and you are supposed
to use configuration files and not whack commands directly.

If 6 months is too much for you on this issue, please feel free to
donate either time or money to the Libreswan Project :)

jlearman:
I couldn't find the source for the documentation.  Point me to that and I might be able to help.

I'd rather not have to spin up a docker just to use `man whack`.  The instance where I run `whack` doesn't (and won't) support `man`.  I don't want to install Libreswan on my dev box for a number of reasons (mostly, so I don't accidentally add a build dependency, as other teammates won't be installing it.)

For applications like k8s Submariner, the `whack` interface is more appropriate than changing config files and poking a daemon.  The file method is ideal for a host or workstation providing a relatively static set of IPsec tunnels, but it's not ideal for multi-cloud interconnection.

cagney:
currently xmlto is used: https://github.com/libreswan/libreswan/blob/main/mk/manpages.mk
sources: https://github.com/libreswan/libreswan/blob/main/programs/pluto/ipsec_pluto.8.xml

cagney:
https://testing.libreswan.org/latest/documentation/
web pages need update.

cagney:
superseeded by nightly documentation builds and

- [ ] #1354

PR #977, opened 2023-01-12
WIP: typos found 2023-01-12

qyanu:
Hi there,

This pull requests includes two typos found, each in their respective commit.

1) One small typo in the pluto manpage is fixed.

2) I notices a warning message in my log like so:
``prepare-client output: vti interface "vti42" already exists with conflicting setting (perhaps need vti-sharing=yes ?``
However, the setting "vti-sharing" does not exist according to the ipsec.conf(5) manpage and a libreswan wiki search. I would say this is a typo, and actually the setting is actually called "vti-shared".

This pull-request is work-in-progress, as i am unsure to what steps to follow to fully test this change (As this does not only affect documentation files).

best regards,
Max

PS: The text in https://libreswan.org/man/ipsec_pluto.8.html points to itself. Is there a publicly available repo to submit a fix?

### DIFF
--- a/programs/_updown.xfrm/_updown.xfrm.in
+++ b/programs/_updown.xfrm/_updown.xfrm.in
@@ -906,7 +906,7 @@ addvtiiface() {
 			echo "vti interface already exists with identical parameters, OK"
 		    fi
 		else
-		    echo "vti interface \"${VTI_IFACE}\" already exists with conflicting setting (perhaps need vti-sharing=yes ?"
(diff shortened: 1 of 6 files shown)


cagney:
> PS: The text in https://libreswan.org/man/ipsec_pluto.8.html points to itself. Is there a publicly available repo to submit a fix?

The file is generated.  Our problem is a lack of a good portable tool for generating the documentation as manual pages and html.  For instance #685 and #747

cagney:
Merged.  Thanks for noticing that the test output needed updating.

PR #986, opened 2023-01-18
IPsec new "Packet" offload support

raed-salem:
The following series aims to add support for the recently added IPsec packet offload as described in the following article:
[Extend XFRM core to allow packet offload configuration](https://lwn.net/Articles/916838/)

in high-level this extends the current crypto nic offload which offloaded the payload crypto operation (encryption/decryption) of an IPsec packet to also offload the IPsec encapsulation (header and ICV) in addition to offload the policies so the entire IPsec data path would be offloaded to supporting nic.

few notes:
- currently no HW supports lifetime limits and so for SA(s) [hard|soft]_[byte|packet]_limit are set to XFRM_INF by default if packet offload is used
- As current implementation does not support socket policies offload (as socket is not necessarily bounded to specific device), there is a need to install an ike port policy which allow IKE traffic to bypass IPsec policies (something similar to poke_ipsec_policy_hole)

Raed Salem (4):
  pluto: nic-offload, add new nic offload type "Packet"
  Add IPsec SA Packet offload
  kernel: Support policy Packet offload
  Kernel: Add IKE policy exception support

 include/pluto_constants.h               |   7 ++
 include/whack.h                         |   2 +-
 lib/libipsecconf/confread.c             |   2 +-
 lib/libipsecconf/keywords.c             |  12 ++-
 programs/pluto/connections.h            |   2 +-
 programs/pluto/iface.c                  |  17 ++++
 programs/pluto/kernel.c                 |  43 +++++++---
 programs/pluto/kernel.h                 |  18 ++++-
 programs/pluto/kernel_policy.c          |  23 +++++-
 programs/pluto/kernel_policy.h          |   1 +
 programs/pluto/kernel_xfrm.c            | 100 ++++++++++++++++++++++--
 programs/pluto/linux-copy/linux/xfrm.h  |   6 ++
 programs/pluto/show_connection_status.c |   5 +-
 programs/whack/whack.c                  |  12 +--
 14 files changed, 213 insertions(+), 37 deletions(-)


paulwouters:
Awesome!
Thanks for the patch. We will have a look and how to deal with the limitations. We will also need to see if the current proposed kernel changes will actually make it in the mainline kernel. I will talk to the XFRM maintainer to see where we are at for that.

lib/libipsecconf/keywords.c
```diff
@@ -114,14 +114,22 @@ static const struct sparse_name kw_yndev_list[] = {
...
+/* Values for yes/no/auto, used by encapsulation */
 static const struct sparse_name kw_yna_list[] = {
 	{ "yes",	yna_yes },
 	{ "no",		yna_no },
 	{ "auto",	yna_auto },
 	SPARSE_NULL
 };
 
+static const struct sparse_name kw_offload_list[] = {
+	{ "yes",        off_yes },
```
paulwouters, line 126: maybe "yes" should be "crypto", so we get:  auto, no, crypto or packet ?

programs/pluto/kernel.c
```diff
@@ -1893,15 +1901,23 @@ static bool setup_half_kernel_state(struct state *st, enum direction direction)
...
+			if (said_next->nic_offload.type == OFFLOAD_PACKET) {
+				said_next->nic_offload.type = OFFLOAD_CRYPTO;
+				ret = kernel_ops_add_sa(said_next, replace, st->st_logger);
+			}
+
+			if (!ret) {
+				/* Fallback to non-nic-offload crypto */
+				said_next->nic_offload.dev = NULL;
+				ret = kernel_ops_add_sa(said_next, replace, st->st_logger);
+			}
```
paulwouters, line 1920: maybe make this conditional on off_auto, and take off_packet and off_yes (off_crypto) to mean only try to do this specific variant (eg to work around unexpected bugs). This way auto could be robust but we could also force to only set crypto or packet offload. This will also give people who want to insist on offload a change to fail the tunnel if they cannot hw offload.

programs/pluto/kernel_xfrm.c
```diff
@@ -1307,10 +1324,18 @@ static bool netlink_add_sa(const struct kernel_state *sa, bool replace,
...
+		req.p.lft.soft_byte_limit = sa->sa_max_soft_bytes;
+		req.p.lft.hard_byte_limit = sa->sa_ipsec_max_bytes;
+		req.p.lft.hard_packet_limit = sa->sa_ipsec_max_packets;
+		req.p.lft.soft_packet_limit = sa->sa_max_soft_packets;
+	} else {
+		req.p.lft.soft_byte_limit = XFRM_INF;
+		req.p.lft.hard_byte_limit = XFRM_INF;
+		req.p.lft.hard_packet_limit = XFRM_INF;
+		req.p.lft.soft_packet_limit = XFRM_INF;
+	}
```
paulwouters, line 1338: If this is a limitation, we should have the connection load already fail on these, except that undermines the usefulness of "auto".

I'm also not sure about this messing with FIPS. We set limited to ensure in-time rekeying but if that is not supported we have to "work around" it with timer based rekeys. I guess for AES that does not matter too much,
since the limits more than 2^32 ? But perhaps at least a warning/debug message should be logged.

What happens when

programs/pluto/kernel_xfrm.c
```diff
@@ -2536,6 +2563,62 @@ static err_t xfrm_migrate_ipsec_sa_is_enabled(struct logger *logger)
...
+
+				.lft.soft_byte_limit = XFRM_INF,
+				.lft.soft_packet_limit = XFRM_INF,
+				.lft.hard_byte_limit = XFRM_INF,
+				.lft.hard_packet_limit = XFRM_INF,
+
+				.sel.proto = IPPROTO_UDP,
+				.sel.family = AF_INET,
+				.sel.sport = htons(IKE_UDP_PORT),
+				.sel.dport = htons(IKE_UDP_PORT),
```
paulwouters, line 2595: what about the NAT port 4500 ? Or is UDP Encap not supported?

programs/pluto/linux-copy/linux/xfrm.h
```diff
@@ -519,6 +519,12 @@ struct xfrm_user_offload {
  */
 #define XFRM_OFFLOAD_IPV6	1
 #define XFRM_OFFLOAD_INBOUND	2
```
paulwouters, line 521: add a comment this requires linux kernel 6.2+

paulwouters (approved):
Looks good in general. Some minor comments and man page updates that need to happen, but we can handle those for you. Give us a few days to merge this in.

paulwouters:
Can you clarify on which hardware this is known to work?

Can I score two cheap ConnectX-2 cards? Or does it requires ConnectX-5 or ConnectX-6 ?

This would be good info to have in the man page (and on our website) too

paulwouters:
also maybe a note on which hardware supports which features / options   (aes128_gcm, aes256_gcm, udp encap, transport|tunnel, ESN, how many keys/tunnels can be supported, etc)

lib/libipsecconf/keywords.c
raed-salem: sounds, good, i thought to keep it as "yes" to be backward compatible

programs/pluto/kernel_xfrm.c
raed-salem: Right i was aware off, UDP Encap is not supported, while at it, shouldn't IKE_UDP_PORT be configurable ? although it is a convention to use 500, however it is not a standard demand to use this port for ike , right ? can you make it configurable if so ?

programs/pluto/linux-copy/linux/xfrm.h
raed-salem: Ack

raed-salem:
> Can you clarify on which hardware this is known to work?
> 
> Can I score two cheap ConnectX-2 cards? Or does it requires ConnectX-5 or ConnectX-6 ?
> 
> This would be good info to have in the man page (and on our website) too

currently starting from ConnectX-7 the feature is supported.

> also maybe a note on which hardware supports which features / options (aes128_gcm, aes256_gcm, udp encap, transport|tunnel, ESN, how many keys/tunnels can be supported, etc)
MLX5 driver currently supports (aes128_gcm, aes256_gcm, transport, keys/tunnels up to 2^24 (SW constrain)

programs/pluto/kernel_xfrm.c
raed-salem: sounds good to me, i guess a debug/warning the way to go, till some supporting HW/implementation is available.

programs/pluto/kernel_xfrm.c
cagney: To follow up Paul's point, shouldn't the ip_port be a parameter so that the code opening sockets can poke multiple holes?  Called from bind_iface_endpoint()?
Bare ESP can be negotiated using any port.  The only got-ya is that the UDP IKE packets sent on ports other than 500 always contain the ESP=0 prefix.

cagney:
> currently starting from ConnectX-7 the feature is supported.

Can KVM model the device?

cagney (changes requested):

programs/pluto/kernel.c
raed-salem: the above code is only if off_auto is set (only then nic_offload_fallback is true)

raed-salem:
> 



> > currently starting from ConnectX-7 the feature is supported.
> 
> Hmm, those are out of range for the libreswan project to purchase. If you can donate two products or engineering samples, we would be able to integrate those into our active daily test harness and do performance and regression testing



> > currently starting from ConnectX-7 the feature is supported.
> 
> Hmm, those are out of range for the libreswan project to purchase. If you can donate two products or engineering samples, we would be able to integrate those into our active daily test harness and do performance and regression testing

i think it can be arranged, please supply me with delivery details and i will try to arrange such cards

raed-salem:
> 

basically we have such, but not sure it is publicly available, will take it internally and report later on

programs/pluto/kernel.c
```diff
@@ -977,7 +978,10 @@ static void revert_kernel_policy(struct spd_route *spd,
 
 	ldbg(logger, "kernel: %s() restoring bare shunt", __func__);
 	struct bare_shunt *bs = *spd->wip.conflicting.shunt;
-	if (!install_bare_shunt_kernel_policy(bs, EXPECT_KERNEL_POLICY_OK,
+	struct nic_offload nic_offload = {};
```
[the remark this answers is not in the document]
raed-salem: i am fine with initializing with {0}.

programs/pluto/kernel_xfrm.c
cagney: IKE UDP packets sent to port 500 don't include the four zero bytes at the very start of the message body, but IKE UDP packets to other ports do.  This only matters when IPsec is being encapsulated within UDP.

Given this card doesn't support UDP encap I'm not sure the zero prefix will matter.  Instead all that's needed is for holes to be punched for all relevant UDP ports.

It's described here https://www.rfc-editor.org/rfc/rfc3948.html

raed-salem: thanks for the reference will look at it, i think i saw this kind of check in XFRM stack also in linux, true IPSEC UDP encap still not supported for offload

paulwouters:
merged into mainline - still needs man page update which i will do today.

PR #994, opened 2023-01-25
Fix referenced before assignment in nsrun script

bradyallenjohnson:
- The TestList __next__() function had 3 variables that were referenced before they were assigned.
- This was encountered on a Fedora 36 cloud-based VM, running the testsuite with Namespaces in the VM. The version of Python on the VM is 3.10.9.

Signed-off-by: Brady Johnson <bradyallenjohnson@gmail.com>


cagney:
@antonyantony merged this

PR #1024, opened 2023-03-01
Use standard ref counting in pluto kernel xfrmi

bradyallenjohnson:
- Remove locally defined ref counting in kernel_xfrm_interface and replace with libswan/refcnt.c
- Also change unclear bool function returns to use standard c-style returns: 0 - success, non-zero - failure

### DIFF
--- a/programs/pluto/connections.c
+++ b/programs/pluto/connections.c
@@ -2343,7 +2343,7 @@ static diag_t extract_connection(const struct whack_message *wm,
 					    wm->xfrm_if_id, err);
 			}
 			if (!setup_xfrm_interface(c, wm->xfrm_if_id == 0 ?
-						  PLUTO_XFRMI_REMAP_IF_ID_ZERO : wm->xfrm_if_id )) {
+						 PLUTO_XFRMI_REMAP_IF_ID_ZERO : wm->xfrm_if_id )) {
 				/* XXX: never happens?!? */
 				return diag("setup xfrmi interface failed");
 			}
--- a/programs/pluto/ikev1_aggr.c
+++ b/programs/pluto/ikev1_aggr.c
@@ -801,7 +801,7 @@ static stf_status aggr_inR1_outI2_crypto_continue(struct state *st,
 	ISAKMP_SA_established(pexpect_ike_sa(st));
 #ifdef USE_XFRM_INTERFACE
 	if (c->xfrmi != NULL && c->xfrmi->if_id != 0)
-		if (add_xfrmi(c, st->st_logger))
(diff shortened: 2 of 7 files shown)


cagney:
The refcnt part looks great.

However:
```
-                       if (!setup_xfrm_interface(c, wm->xfrm_if_id == 0 ?
-                                                 PLUTO_XFRMI_REMAP_IF_ID_ZERO : wm->xfrm_if_id )) {
+                       if (setup_xfrm_interface(c, wm->xfrm_if_id == 0 ?
+                                                PLUTO_XFRMI_REMAP_IF_ID_ZERO : wm->xfrm_if_id )) {
                                /* XXX: never happens?!? */
                                return diag("setup xfrmi interface failed");
```
libreswan's code base mostly tries to steer clear of this style.  For instance, string routines use streq() et.al. wrappers.  Exceptions tend to get flagged with:
```
-/* ??? true means failure! */
```

cagney:
FYI, the const cast shouldn't be needed:
```
+                       refcnt_peek((const struct refcnt *) &(c->xfrmi->refcnt)));
```

cagney:
merged by cd6326d638a2cb1bfbfcce33e504452f4a5bb8e6

cagney:
fyi, testing found a leak, https://testing.libreswan.org/?run=v4.9-995-gcd6326d638-main&run=v4.9-990-g99c52cb4c7-main

cagney:
part of the problem is the (forced) use of the global logger, I'll push change so that the connection's logger is passed down

cagney:
I'm wondering about this:
```
@@ -418,7 +418,7 @@ static void process_nlmsgs(char *msgbuf,  ssize_t len, struct ifinfo_response *i
  i++;
  red_msg_size += nlmsg->nlmsg_len;
  dbg("RTM_NEWLINK: netlink %d ignored %d. Bytes %d", i, ignored, red_msg_size);
- if (parse_nl_newlink_msg(nlmsg, ifi_rsp) == 0 && ifi_rsp->result)
+ if (parse_nl_newlink_msg(nlmsg, ifi_rsp) != XFRMI_SUCCESS && ifi_rsp->result)
  return;
  break;
 ```

cagney:
closer https://testing.libreswan.org/v4.9-1016-g9570e97c55-main/ikev2-xfrmi-02/OUTPUT/road.console.diff

cagney:
that fixed the last regression

bradyallenjohnson:
Andrew,

Sorry for not working on this yesterday or today. It's been one of those
weeks and it's only Tuesday...

Thanks for fixing this, I'll look at the details tomorrow.

Regards,

Brady

PR #1105, opened 2023-04-22
testing: remove eternal WIP test that's about unsupported PPK feature…

vukasink:
… option


cagney:
coming soon; resolving WIP tests is always good :-)
(I also removed the TESTLIST entry)

PR #1128, opened 2023-05-07
Introduce ppk-ids connection option as a prep work for draft-smyslov-ipsecme-ikev2-qr-alt-07 (PPKs during INTERMEDIATE exchange)

vukasink:
ppk-ids option allows for specifying a list of PPK_IDs to look for in .secrets file

I'm open to suggestions if something needs to be improved in the patch.

### DIFF
--- a/configs/Makefile
+++ b/configs/Makefile
@@ -153,6 +153,7 @@ XMLSOURCES += d.ipsec.conf/ms-dh-downgrade.xml
 XMLSOURCES += d.ipsec.conf/dns-match-id.xml
 XMLSOURCES += d.ipsec.conf/require-id-on-certificate.xml
 XMLSOURCES += d.ipsec.conf/ppk.xml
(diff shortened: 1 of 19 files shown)


vukasink:
Anyone had a change to take a look at this? :)

cagney:
FYI, DBG_log() always debug-logs so ldbg(c->logger, ...) for many cases will be better

cagney:
I see both get_connection_ppk_initiator() and ..._responder() slice and dice c->config->ppk_ids; should that instead be done once when adding the connection?

vukasink:
Thanks, I'll update the code to use d605ba5648

vukasink:
Updated the code to use d605ba5.

At some point I had to turn shunk_t into chunk_t by casting away the const qualifier.

An alternative would be to clone shunk_t into a new chunk_t, and then free the newly created chunk in the caller function, which seemed more complicated approach to me. But I do not have a strong opinion about it, so if anybody objects, let me know.

PR #1251, opened 2023-08-27
Reference count .st_connection

cagney:
The missing bit of reference counting connections.
100% passes testsuite.


cagney:
merged

PR #1320, opened 2023-10-09
Add option for the dont-encap-dscp xfrm flag

wofferl:
Use the xfrm dont-encap-dscp flag to prevent TOS/DSCP flags from being copied to the encrypted packet. This is the equivalent to the old hidetos option with klips.

   **hidetos**
           _whether a tunnel packet's TOS field should be set to 0 rather than copied from the user packet inside; acceptable values are yes (the default) and no. This KLIPS option has no effect on NETKEY, Windows or BSD stacks._


bleve:
It is very important that options are positive. So options should always be enable something, never disable something. That is because disable=no is very, very confusing.

wofferl:
Ah ok I see. I will change this.
And the default? 
I think as it is currently, which would mean encap-dscp=yes.

bleve:
Sounds good.

cagney:
So we've got:
`encap-dscp=` is for the outgoing path - default yes thanks to the kernel
`decap-dscp=` is for the inbound path - default no thanks to the kernel
I'll  stick with the internal field name .encap_dscp. as  .dont_encap_dscp leads to !.dont_encap_dscp.
And like you've done, add DONT_ENCAP_DSCP in the policy (I think it should be moved out of policy, but that is something post merge).

cagney:
merged

cagney:
ps: @wofferl you must have been reading the code pretty carefully as you managed to use stuff I only tweaked over the weekend!

PR #1409, opened 2023-11-16
Iptfs aa 20231113

antonyantony:
I'm sharing a proof of concept version that I've been using to test kernel patches. 
I believe it's most appropriate to model user interface identical to (kernel/iproute2) use, essentially treating it as a new 'mode' called iptfs. You can argue that :) bring that to kernel patch RFC going around:)

Additionally, I envision optimizing my current version "type=tunnel|transport|iptfs" approach by offering a simple option for users in the next iteration, which would be 'type=tunnel|transport.'

Under the hood, 'tunnel' would correspond to IP-TFS or tunnel mode. I plan to explore this further in future iterations.

libreswan related feedback here  and kernel patches sent to nedev or devel-ipsec.

Kernel deteection is incomplete, there is a segfault. That is why reverted the last commit. I will work on that later. I am also wondering to create policy or state for detecting. If it is state I don"t have to delete it, instead create it with 1 sec hard expire and it will disappear on its own:)


cagney:
While I so wish we had `mode={transport,tunnel,ip*tfs}` because of the way the RFCs are written (for instance, not negotiating transport should create a tunnel), I susepect we'll end up with transport=yes|no|must tptfs=... et.al.  But that is something to solve after this is merged.

antonyantony:
The fallback from Transport to Tunnel, RFC 7296, is hard to understand for me, especially from a user's perspective. If the user specification requests Transport mode and then fallback to Tunnel mode could be confusing. Additionally, consider scenarios in which hardware offload exclusively supports Transport Mode; this could lead to suboptimal performance if it's forced to fall back to Tunnel mode.

I remember once mentioning fallback to Tunnel mode is requirement to a hardware vendor that only supported Transport mode, there was silence in the room :) 
I'm curious about how widespread the support and usage of IKEv2 fallback from Transport to Tunnel mode is. 

Let's see if we can simplify it to 'pluto's type=tunnel|iptfs|transport.' Even with separate keywords, propose IP-TFS and fallback to Tunnel might complicate matters. Otherwise end up with ikev1 leave it with type=tunnel|transport and
ikev2 got to what you propose. transport=yes|no|must.

cagney:
Yes, `type=tunnel|iptfs|transport|...` is enough for now.

antonyantony:
detecting IPTFS support is working now. Though need fixing to able to merge to main.

cagney:
```

static bool qry_xfrm_iptfs_support(const struct kernel_state *sa, struct logger *logger)
{

	return netlink_add_sa(sa, false, logger);
}
```
should it also call `xfrm_del_ipsec_spi()`.  I do like how high level code fakes up the SA making this code simpler.

cagney:
```
	ip_address loopback;
	char loopback_addr[] = "127.0.0.1";
	const struct ip_info *type = NULL;
	err_t err = ttoaddress_num(shunk1(loopback_addr), type, &loopback);
```
fyi, loopback is hiding in ipv4_info.address.loopback

cagney:
check `ikev2-12-transport-psk`, I suspect the code:
```
	if (child->sa.st_esp.present) {
		policy.esp = true;
		mode = (child->sa.st_esp.attrs.mode == ENCAPSULATION_MODE_TUNNEL ? ENCAP_MODE_TUNNEL :
			child->sa.st_esp.attrs.mode == ENCAPSULATION_MODE_IPTFS ? ENCAP_MODE_IPTFS :
			pexpect(0));
	}
```
needs to be aware of transport mode vis:
```
| kernel: install_inbound_ipsec_kernel_policies() is installing SPD for 192.1.2.45/32=>192.1.2.23/32
EXPECTATION FAILED: 0 (kernel_policy_from_state() +245 programs/pluto/kernel_policy.c)
| kernel: NIC esp-hw-offload not for connection 'ipv4-psk-ikev2-transport' not available on interface eth1
FATAL: ASSERTION FAILED: switch (mode) case 0 (0x0) unexpected (kernel_policy_from_spd() +147 programs/pluto/kernel_policy.c)
```

cagney:
this gets around an assumption that things are ESP, only to replace it with a assumption that the connection is IP-TFS only
```
diff --git a/programs/pluto/ikev2_child.c b/programs/pluto/ikev2_child.c
index e8ecc73435..e4215c7386 100644
--- a/programs/pluto/ikev2_child.c
+++ b/programs/pluto/ikev2_child.c
@@ -650,8 +650,7 @@ void llog_v2_child_sa_established(struct ike_sa *ike UNUSED, struct child_sa *ch
                jam_so(buf, child->sa.st_clonedfrom);
                jam_string(buf, "; IPsec ");
                /* log Child SA Traffic Selector details for admin's pleasure */
-               jam_string_human(buf, enum_name_short(&encapsulation_mode_names,
-                                                     child->sa.st_esp.attrs.mode));
+               jam_enum_human(buf, &encap_mode_names, c->config->child_sa.encap_mode);
                FOR_EACH_ITEM(spd, &c->child.spds) {
                        jam_string(buf, " ");
                        jam_string(buf, "[");
```

cagney:
fyi,
```
--- a/programs/pluto/connections.c
+++ b/programs/pluto/connections.c
@@ -2134,20 +2134,6 @@ static diag_t extract_connection(const struct whack_message *wm,
                }
        }
 
-       if (mode == ENCAP_MODE_TUNNEL) {
-               if (kernel_ops->migrate_ipsec_sa_is_enabled == NULL) {
-                       return diag("MOBIKE is not supported by %s kernel interface",
-                                       kernel_ops->interface_name);
-               }
-               /* probe the interface */
-               err_t err = kernel_ops->migrate_ipsec_sa_is_enabled(c->logger);
-               if (err != NULL) {
-                       return diag("MOBIKE support is not enabled for %s kernel interface: %s",
-                                       kernel_ops->interface_name, err);
-               }
-
-       }
-
        if (mode == ENCAP_MODE_IPTFS) {
                if (kernel_ops->iptfs_is_enabled == NULL) {
                        return diag("IPTFS is not supported by %s kernel interface",
```
there's already code further down check for mobike - the above trips on bsd where support is missing

antonyantony:
> ```
> 	ip_address loopback;
> 	char loopback_addr[] = "127.0.0.1";
> 	const struct ip_info *type = NULL;
> 	err_t err = ttoaddress_num(shunk1(loopback_addr), type, &loopback);
> ```
> 
> fyi, loopback is hiding in ipv4_info.address.loopback

I am re-writing this function. There is an easy way, by checking the proc value.

cagney:
Mainline, unfortunately, ended up with something different.

PR #1465, opened 2023-12-14
Fix make git-rpm

bradyallenjohnson:
- It was broken by a previous commit that uncommented "global prever" in packaging/rhel/8/libreswan.spec, since the "make git-rpm" script was expecting it to be commented.


cagney:
87956ac7196

cagney:
merged

PR #1467, opened 2023-12-14
terminate: terminate_and_down_connections() fixes

cagney:
- assume/make caller hold a reference

  otherwise terminate_and_down_connections() can delete the caller's CK_INSTANCE, CK_LABELED_PARENT

- don't allow CK_LABELED_CHILD

  can only meaningfully down the parent or template

- for CK_INSTANCE and CK_PERMANENT check pending and revival, and unroute

  can't, for instance, assume an unrouted connection isn't waiting on another connection (pending), or revival

### DIFF
--- a/programs/pluto/foodgroups.c
+++ b/programs/pluto/foodgroups.c
@@ -111,7 +111,7 @@ static void delete_group_instantiation(co_serial_t serialno, struct logger *logg
 	connection_attach(template, logger);
 	ldbg(template->logger, "removing group template");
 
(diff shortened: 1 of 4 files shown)


cagney:
`terminate_and_down_connections()` is called by `orient()` when it finds that there's some sort of conflict and the existing connection should be terminated.  For instance, during revival, the interfaces are changed so that a connection is no longer oriented.

The easier way to test it is to modify `ipsec shutdown` to use this function (not part of this patch).

cagney:
merged

PR #1482, opened 2023-12-16
building openbsd: don't try to use SADB_X_SPDFLUSH

cagney:
On OpenBSD, SADB_FLUSH flushes everything.


cagney:
tundra got further with building OpenBSD than testing

cagney:
merged

PR #1483, opened 2023-12-16
building: fix typo in ok[] initializer

cagney:
bizarely only OpenBSD saw this as a problem!


cagney:
6a91a32c3d44c64c899d1c40938555b485e0d76c

PR #1491, opened 2023-12-18
connections: dump the connection instance that is still using template

cagney:
In delete_connection(), when deleting a template or group, dump the details of an instances (they should not be there)


cagney:
This changes logging when there's an invalid conn-delete(); results were clean; but waiting on a fresh run using top-of-tree

cagney:
merged

PR #1511, opened 2024-01-03
pluto: handle partial install_inbound_ipsec_kernel_policy() failures

paulwouters:
This was assumed to never fail, but can fail for various reasons, including trying to use hardware offload that does not support the current properties of the IPsec SA.

eg it could install the "in" policy, then try the "fwd" policy and fail. But it would continue doing the "out" policy and then claim successful IPsec SA.

When a failure happens, try to cleanup partially installed IPsec SA.

### DIFF
--- a/programs/pluto/kernel.c
+++ b/programs/pluto/kernel.c
@@ -1714,7 +1714,13 @@ static bool install_inbound_ipsec_kernel_policies(struct child_sa *child)
 		     /* inbound */
 		     str_selector(&spd->remote->client, &sb),
 		     str_selector(&spd->local->client, &db));
-		install_inbound_ipsec_kernel_policy(child, spd, HERE);
+		if (!install_inbound_ipsec_kernel_policy(child, spd, HERE)) {
+		    log_state(RC_LOG_SERIOUS, &child->sa, "Installing IPsec SA failed - check logs or dmesg");
+		    struct spd_owner owner = spd_owner(spd, RT_UNROUTED_FAILURE, logger, HERE);
+		    delete_spd_kernel_policies(spd, &owner, IGNORE_KERNEL_POLICY_MISSING,
+						logger, HERE, "cleanup after failed add_sa");
+			return false;
+		}
 	}
 
 	return true;
--- a/programs/pluto/kernel_policy.c
+++ b/programs/pluto/kernel_policy.c
@@ -576,7 +576,7 @@ static bool pexpect_cat(const struct connection *c, struct logger *logger)
 		PEXPECT(logger, c->local->child.has_cat));
 }
 
-void add_cat_kernel_policy(const struct connection *c,
+bool add_cat_kernel_policy(const struct connection *c,
 			   const struct kernel_policy *kernel_policy,
 			   enum direction direction,
 			   struct logger *logger, where_t where,
@@ -584,7 +584,7 @@ void add_cat_kernel_policy(const struct connection *c,
 {
 	ldbg(logger, "%s", reason);
 	if (!pexpect_cat(c, logger)) {
-		return;
+		return false;
 	}
 
 	ip_selector local_client = selector_from_address(kernel_policy->local.host);
@@ -593,7 +593,9 @@ void add_cat_kernel_policy(const struct connection *c,
 			       kernel_policy, deltatime(0),
 			       logger, where, reason)) {
 		llog(RC_LOG, logger, "%s failed", reason);
+		return false;
 	}
+	return true;
 }
 
 static void delete_cat_kernel_policy(const struct spd_route *spd,
@@ -656,7 +658,7 @@ void delete_cat_kernel_policies(const struct spd_route *spd,
 	}
 }
 
-void install_inbound_ipsec_kernel_policy(struct child_sa *child,
+bool install_inbound_ipsec_kernel_policy(struct child_sa *child,
 					 struct spd_route *spd,
 					 where_t where)
 {
@@ -677,9 +679,11 @@ void install_inbound_ipsec_kernel_policy(struct child_sa *child,
 	const char *add_inbound_cat = NULL;
 #endif
 	if (add_inbound_cat != NULL) {
-		add_cat_kernel_policy(child->sa.st_connection,
+		if(!add_cat_kernel_policy(child->sa.st_connection,
 				      &kernel_policy, DIRECTION_INBOUND,
-				      child->sa.logger, where, add_inbound_cat);
+				      child->sa.logger, where, add_inbound_cat)) {
+			return false;
+		}
 	}
 
 	if (!kernel_ops_policy_add(KERNEL_POLICY_OP_ADD,
@@ -695,7 +699,9 @@ void install_inbound_ipsec_kernel_policy(struct child_sa *child,
 			__func__,
 			str_selector(&kernel_policy.src.client, &sb),
 			str_selector(&kernel_policy.dst.client, &db));
+		return false;
 	}
+	return true;
 }
 
 bool install_outbound_ipsec_kernel_policy(struct child_sa *child,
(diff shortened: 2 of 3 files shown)


cagney:
This is similar to what `revert_kernel_policy()` is trying to do.  Except, unlike this patch, it goes through all the SPDs trying to undo the damage.

However I'm pretty sure `revert_kernel_policy()` is also broken as it assumes that the right thing to do is to back out (probably delete) the policy  What could be needed is a blocking or on-demand policy but this function is too low-level to make such high level decision.

Which makes for more questions:

- how did we let things get this far
For instance, if hardware doesn't allow a configuration then why was it loaded.  Better to fix that.
- what should happen next
If the kernel rejects what was installed, should the connection retry or be disabled?
- should we be aiming for anything approaching perfection
Would it be easier to just pretend everything worked, and then transition to the next state ignoring errors.

letoams:
On Jan 7, 2024, at 12:33, cagney ***@***.***> wrote:
> 
> 
> This is similar to what revert_kernel_policy() is trying to do. Except, unlike this patch, it goes through all the SPDs trying to undo the damage.
> 

Oh. We could try using that instead.

> However I'm pretty sure revert_kernel_policy() is also broken as it assumes that the right thing to do is to back out (probably delete) the policy What could be needed is a blocking or on-demand policy but this function is too low-level to make such high level decision.
> 

Yeah.

> Which makes for more questions:
> 
> how did we let things get this far
> For instance, if hardware doesn't allow a configuration then why was it loaded. Better to fix that.
That’s the other PR. But I wanted something here to ensure failure. Without this patch, pluto thought the IPsec SA was installed correctly.
Better fail bad than cause clear text leak.

> what should happen next
> If the kernel rejects what was installed, should the connection retry or be disabled?
Like other failures (eg auth fail), it should keep
retrying with exponential back off.

> should we be aiming for anything approaching perfection
> Would it be easier to just pretend everything worked, and then transition to the next state ignoring errors.

What is “next state”?

I thought of adding a lower priority software block shunt when doing a hardware SA, so we always have a block in place in case the hardware failed but this would be a new kind of shunt and I didn’t want to introduce that at this point. I was also not sure how it would interact with hardware and performance.

cagney:
> That’s the other PR. But I wanted something here to ensure failure. Without this patch, pluto thought the IPsec SA was installed correctly.

Is that still true?

I pushed the seemingly trivial but powerful 8b0d14c0ec0db2d99a017bca9dea3076326500ce along with the tests 3bdebc011cf4eaced7d0ec455326617d131c3bc0 to get the routing state machine back on track.

You might try changing code like:
```
                if (!install_inbound_ipsec_sa((*e->child), RT_UNROUTED_INBOUND, e->where)) {
                        return false;
                }
                set_routing(event, c, RT_UNROUTED_INBOUND, e);
```
to:
```
                if (!install_inbound_ipsec_sa((*e->child), RT_UNROUTED_INBOUND, e->where)) {
                        /* force transition so next transition undoes it? */
                        set_routing(event, c, RT_UNROUTED_INBOUND, e);
                        return false;
                }
                set_routing(event, c, RT_UNROUTED_INBOUND, e);
```
and see what effect it has on those tests.

cagney:
>> what should happen next
If the kernel rejects what was installed, should the connection retry or be disabled?

> Like other failures (eg auth fail), it should keep
retrying with exponential back off.

But this is a h/w failure.  Without some sort of intervention it will keep failing.  If hw-offload is auto, should it try failing back to software; or even do that inside the kernel and not fail the call?

letoams:
On Mon, 8 Jan 2024, cagney wrote:

>             what should happen next
>             If the kernel rejects what was installed, should the connection retry or be disabled?
>
>       Like other failures (eg auth fail), it should keep
>       retrying with exponential back off.
> 
> But this is a h/w failure.

Why does that matter?

> Without some sort of intervention it will keep failing.

Yes, same as authentication failure. or wrong TS. or other failures.

> If hw-offload is auto, should it try failing back to software; or even
> do that inside the kernel and not fail the call?

It already does that at the higher level:

                 bool ret = kernel_ops_add_sa(said_next, replace, st->logger);

                 if (!ret && nic_offload_fallback &&
                         said_next->nic_offload.dev != NULL) {
                         /* Fallback to crypto offload from packet offload */
                         if (said_next->nic_offload.type == OFFLOAD_PACKET) {
                                 said_next->nic_offload.type = OFFLOAD_CRYPTO;
                                 ret = kernel_ops_add_sa(said_next, replace, st->logger);
                         }

                         if (!ret) {
                                 /* Fallback to non-nic-offload crypto */
                                 said_next->nic_offload.dev = NULL;
                                 ret = kernel_ops_add_sa(said_next, replace, st->logger);
                         }

eg if we configured tunnel mode with nic-offload=packet:

***@***.***:~# ipsec up test
"test" #1: initiating IKEv2 connection to 10.0.1.2 using UDP
"test" #1: sent IKE_SA_INIT request to 10.0.1.2:500
"test" #1: sent IKE_AUTH request {cipher=AES_GCM_16_128 integ=n/a prf=HMAC_SHA2_512 group=MODP2048}
"test" #1: STATE_V2_PARENT_I2: retransmission; will wait 0.5 seconds for response
"test" #1: initiator established IKE SA; authenticated peer using authby=secret and ID_IPV4_ADDR '10.0.1.2'
"test" #2: initiator established Child SA using #1; IPsec tunnel [10.0.1.1/32===10.0.1.2/32] {ESP/ESN=>0x5f65f430 <0x0fee4845 xfrm=AES_GCM_16_128-NONE DPD=passive}
***@***.***:~# ip xfrm state |grep offload
 	crypto offload parameters: dev ens8191f0np0 dir out mode crypto
 	crypto offload parameters: dev ens8191f0np0 dir in mode crypto

(our logging on whack socket needs improving though)

Paul

cagney:
>> But this is a h/w failure.

> Why does that matter?

>> Without some sort of intervention it will keep failing.

> Yes, same as authentication failure. or wrong TS. or other failures.

The two scenarios are not the same:

- when authentication fails or TS is rejected, a connection routing transition is triggered
at all times the connection is in a well defined state
- these should-not-happen failing kernel calls occur mid-transtion
the connection no longer has a well defined state

>> If hw-offload is auto, should it try failing back to software; or even do that inside the kernel and not fail the call?

> It already does that at the higher level:

Not always, OFFLOAD_PACKET only falls back to OFFLOAD_CRYPTO, not software?  Is there anything in connectionstatus to show that this happened?

paulwouters:
note set_routing() is static so it cannot be used in kernel.c.

cagney:
I guess the immediate thing to do is add the code that propagates the failure up the stack but hold off on any attempts at recovery.

It means that this:
```
		if (!install_inbound_ipsec_kernel_policy(child, spd, HERE)) {
		    log_state(RC_LOG_SERIOUS, &child->sa, "Installing IPsec SA failed - check logs or dmesg");
		    struct spd_owner owner = spd_owner(spd, RT_UNROUTED_FAILURE, logger, HERE);
		    delete_spd_kernel_policies(spd, &owner, IGNORE_KERNEL_POLICY_MISSING,
						logger, HERE, "cleanup after failed add_sa");
			return false;
		}
```
is reduced to this:
```
		if (!install_inbound_ipsec_kernel_policy(child, spd, HERE)) {
			return false;
		}
```
for now (install_outbound_ipsec_kernel_policy() should have already logged).  Hopefully this causes no regressions and the impair-install-inbound-ipsec-sa* tests don't start core dumping again.

Currently --impair install-*-ipsec-sa induces a late failure (after everything is done), we'll want to add a way to induce an early failure to see its effect.

cagney:
> ```
>                 if (!install_inbound_ipsec_sa((*e->child), RT_UNROUTED_INBOUND, e->where)) {
>                         /* force transition so next transition undoes it? */
>                         set_routing(event, c, RT_UNROUTED_INBOUND, e);
>                         return false;
>                 }
>                 set_routing(event, c, RT_UNROUTED_INBOUND, e);
> ```
> 
> and see what effect it has on those tests.

My initial test shows a weird PEXPECT fail, investigating.

paulwouters:
Only the failure path up the stack implemented. No remedies are implemented. Also nic-offload= no longer supports fallback. New code pushed via different commit, obsoleting this PR.

PR #1512, opened 2024-01-03
pluto: check various incompatible settings with nic-offload=packet|auto

paulwouters:
- Limit the replay-window size to what is supported in known HW. (but what to do with replay-window=0 and it disabling ESN?)
- Only allow ESP, not AH or IPTFS
- Do not allow compression
- TODO: what about  tfcpad= , encap-dscp, nopmtudisc, ikepad, encapsulation,

### DIFF
--- a/programs/pluto/connections.c
+++ b/programs/pluto/connections.c
@@ -2140,9 +2140,18 @@ static diag_t extract_connection(const struct whack_message *wm,
 	config->child_sa.encap_mode = mode;
 
 	if (mode == ENCAP_MODE_TRANSPORT) {
-		if (wm->vti_interface != NULL) {
+		if (wm->vti_interface != NULL)
 			return diag("VTI requires tunnel mode but connection specifies type=transport");
-		}
+	} else {
+		if (wm->nic_offload == offload_packet || wm->nic_offload == offload_auto)
+			return diag("nic-offload=packet|auto currently works for transport mode only. There are no known nic-offload cards that support packet offload with Tunnel Mode or IPTFS. If you know of such a card, contact swan-dev@lists.libreswan.org");
+	}
+
+	if (wm->nic_offload == offload_packet || wm->nic_offload == offload_auto) {
+		if (encap_proto != ENCAP_PROTO_ESP)
+			return diag("nic-offload=packet|auto requires phase2=esp");
+		if (compress)
+			return diag("nic-offload=packet|auto does not support IPCOMP compression.");
 	}
 
 	if (wm->authby.never) {
@@ -2572,6 +2581,20 @@ static diag_t extract_connection(const struct whack_message *wm,
 			    kernel_ops->max_replay_window);
 	} else if (!never_negotiate_wm(wm)) {
 		config->child_sa.replay_window = wm->replay_window;
+	}
+
+	if (wm->nic_offload == offload_packet || wm->nic_offload == offload_auto) {
+		switch (wm->replay_window) {
+			case 32:
+			case 64:
+			case 128:
+			case 256:
+				dbg("packet offload replay-window compatible with all known hardware and Linux kernels");
+				break;
+			default:
+				return diag("current packet offload hardware only supports replay-window of 32, 64, 128 or 256");
+		}
+		/* check if we need checks for tfcpad= , encap-dscp, nopmtudisc, ikepad, encapsulation, etc? */
 	}
 
 	if (never_negotiate_wm(wm)) {


bleve:
Looks good.

about replay-window=0 we can't do anything - what happens in that case is defined by rfc, only thing we might do is give warning, but warning like "Warning: reply-window=0 (disable) also disables ESN, not suitable for fast transfer speeds"

cagney:
Why not probe the hardware.

letoams:
On Jan 7, 2024, at 12:35, cagney ***@***.***> wrote:
> 
> Why not probe the hardware.
> 
Combinatory explosion?
Not guaranteed the hardware throws an error?

It’s worth doing this, once we see more hardware implementations. This is indeed a bit of a bandaid. Perhaps there will be proper features discovery call in Linux in the future ?

Note also technically, you might not know over which nic a future IPsec tunnel would go over, so you might not know which interface probe answers to use.

Paul

cagney:
>> Why not probe the hardware.

Or at least the kernel.[hc].  Adding:
```
dbg("packet offload replay-window compatible with all known hardware and Linux kernels");
```
to generic code should, minimally, raise an eyebrow.

> Combinatory explosion?
> Not guaranteed the hardware throws an error?

> It’s worth doing this, once we see more hardware implementations. This is indeed a bit of a bandaid. Perhaps there will be proper features discovery call in Linux in the future ?

> Note also technically, you might not know over which nic a future IPsec tunnel would go over, so you might not know which interface probe answers to use.

A valid point.  Pluto already handles this though with per-interface hw detection vis:
```
ifd->nic_offload = kernel_ops_detect_offload(ifp, logger);
```
However, a grep of where that field is referenced only turns up kernel*.c - shouldn't there be a check in orient() so that a connection with hw-offload=yes only picks those interfaces.  Ditto for the above?

cagney:
BTW
>  Ditto for the above?
isn't copy the code; rather orient ask the kernel if the interface can support offload with a given set of parameters

cagney:
Should this be toned down?  Warn, rather than reject, combinations that are considered suspect on known hardware.

programs/pluto/connections.c
```diff
@@ -535,886 +489,3124 @@ static void discard_connection(struct connection **cp, bool connection_valid, wh
...
+		if (wm->nic_offload == offload_packet || wm->nic_offload == offload_auto)
+			return diag("nic-offload=packet|auto currently works for transport mode only. There are no known nic-offload cards that support packet offload with Tunnel Mode or IPTFS. If you know of such a card, contact swan-dev@lists.libreswan.org");
+	}
+
+	if (wm->nic_offload == offload_packet || wm->nic_offload == offload_auto) {
+		if (encap_proto != ENCAP_PROTO_ESP)
+			return diag("nic-offload=packet|auto requires phase2=esp");
+		if (compress)
+			return diag("nic-offload=packet|auto does not support IPCOMP compression.");
+	}
```
cagney, line 2155: this is merged with:
```
	if (nic_offload == NIC_OFFLOAD_PACKET) {
		if (encap_mode != ENCAP_MODE_TRANSPORT) {
			return diag("nic-offload=packet restricted to type=transport");
		}
		if (encap_proto != ENCAP_PROTO_ESP) {
			return diag("nic-offload=packet restricted to phase2=esp");
		}
		if (compress) {
			return diag("nic-offload=packet restricted to compress=no");
		}
		if (config->encapsulation == YNA_YES) {
			return diag("nic-offload=packet cannot specify encapsulation=yes");
		}
```

programs/pluto/connections.c
```diff
@@ -535,886 +489,3124 @@ static void discard_connection(struct connection **cp, bool connection_valid, wh
...
+			case 128:
+			case 256:
+				dbg("packet offload replay-window compatible with all known hardware and Linux kernels");
+				break;
+			default:
+				return diag("current packet offload hardware only supports replay-window of 32, 64, 128 or 256");
+		}
+		/* check if we need checks for tfcpad= , encap-dscp, nopmtudisc, ikepad, encapsulation, etc? */
+	}
+
```
cagney, line 2599: this is merged with
```
	if (nic_offload == NIC_OFFLOAD_PACKET) {
		/* limited replay windows supported for packet offload */
		switch (replay_window) {
		case 32:
		case 64:
		case 128:
		case 256:
			vdbg("packet offload replay-window compatible with all known hardware and Linux kernels");
			break;
		default:
			return diag("current packet offload hardware only supports replay-window of 32, 64, 128 or 256");
		}
```

cagney:
equivalent code merged

PR #1646, opened 2024-03-12
fix spelling

dkg:
This clears up the overwhelming majority of complaints produced by the following command:

    codespell --ignore-words-list bu,ot,dran,re-use,gir,racoon,uptodate,wan,nd,ue,ba,te,ans,ure,cas,crasher,keep-alives,oen,gost,keypairs,crashers,parms,hist,keypair,ser --skip ./docs/CHANGES.openswan,./LIBRESWAN-GPG-KEY.txt,./packaging/d*/copyright,./packaging/d*/upstream/signing-key.asc,./security/CVE-*.txt

It might make sense to include something like this command, with an embedded ignore-words-list and skiplist, in the repository so that it becomes easier to do an automated codespell as part of CI or as a pre-commit hook.


dkg:
If this is merged, and i hear any interest from the libreswan developers, i will consider offering a PR that includes a `make spellcheck` target, which you can integrate into your CI workflow to keep the codebase orthographically clean.

I know these things are trivial compared to functionality concerns, but misspelled documentation and error messages cast some amount of doubt on the attention to detail in the codebase.  And i'd prefer that kind of skepticism to be reserved for more meaningful errors.

cagney:
Sure ....
I assume there's a Fedora package.  We can add it to the table at the top of https://testing.libreswan.org/current/ (which I see is all ok!).

dkg:
OK, this should now merge cleanly again.

dkg:
i keep updating this to match the main branch and clean up minor conflicts.  it'd be great if it could get a review (even better if that results in a merge) so i don't have to keep doing that. ☺

cagney:
first part is merged; running testsuite

cagney:
merged

PR #1647, opened 2024-03-12
fix trivial orthography on 4.x maintenance branch

dkg:
This is not as comprehensive as #1646 (which is focused on the main 5.x development branch) , but it clears up a few spelling errors in the 4.x series that are automatically detected by lintian's automated spellchecking of strings and manpages.


paulwouters:
I have applied there, but of course we hope to not do any 4.x releases anymore. But if we do, these will be part of it. Thanks!

PR #1674, opened 2024-04-03
Use reference implementation of sd_notify and remove dep on libsystemd.

gordonmessmer:


cagney:
Unfortunately I couldn't get this to build.  We've pushed an alternative into mainline.

PR #1689, opened 2024-04-25
Suppress warnings with GCC 14

ueno:
This fixes the following compiler warning when building on Fedora rawhide (Fedora 41):
```console
  parser.lex: In function ‘glob_include’:
  parser.lex:198:44: error: ‘calloc’ sizes specified with ‘sizeof’ in the earlier argument and not in the later argument [-Werror=calloc-transposed-args]
    198 |         stacktop->fileglob = calloc(sizeof(char *), count + 1);
        |                                            ^~~~
  parser.lex:198:44: note: earlier argument should specify number of elements, later size of each element
```

### DIFF
--- a/lib/libipsecconf/parser.lex
+++ b/lib/libipsecconf/parser.lex
@@ -46,6 +46,7 @@ YY_DECL;
 #include "ipsecconf/parserlast.h"
 #include "lswlog.h"
 #include "lswglob.h"
+#include "lswalloc.h"
 
 #define MAX_INCLUDE_DEPTH	10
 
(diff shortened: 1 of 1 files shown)


cagney:
backstory https://gcc.gnu.org/pipermail/gcc-patches/2023-December/641030.html

ueno:
> backstory https://gcc.gnu.org/pipermail/gcc-patches/2023-December/641030.html

I see. By the way, this is the only place where bare `calloc` is used. Maybe it could be replaced with `alloc_things`?

cagney:
> I see. By the way, this is the only place where bare `calloc` is used. Maybe it could be replaced with `alloc_things`?

Very Good point.  The corresponding free() call, at least, would also need to be updated.  And a check on where it is used so warrants at least a sniff test.

cagney:
Nice, I'll merge it once the branches are unfrozen.

cagney:
btw, only calloc() in 4.x looks like `calloc(1, size)`.  Anyway testing.

cagney:
19eb4a08e400e05167aed2201222be4b1c18fd3a merged!

PR #1698, opened 2024-05-02
add xfrm interface before routing connection

wofferl:
Currently, the XFRM interface is created when the tunnel is up.  With this patch the interface is also created before routing a connection with ipsec route or auto=route.


cagney:
running a full testsuite

cagney:
merged

PR #1703, opened 2024-05-06
xfrmi: prevent routes from multiple tunnels that are still in use from being deleted

wofferl:
It actual can happen when using multiple tunnels to the same peer with xfrmi and mark-out, that the xfrmi table route is deleted by one tunnel, making the other tunnel non-functional.

However, this requires one tunnel to be configured directly to the gateway.

```
192.1.3.33 <==> 192.1.2.23      from 192.1.3.33 to 192.1.2.23 (84B/84B) north-east-gw, reqid=16388
192.1.3.33 <==> 192.0.21.0/24   from 192.1.3.33 to 192.1.2.23 (84B/84B) north-east-sn, reqid=16392
```

If the subnet tunnel is terminated first, the gateway tunnel cannot be terminated cleanly, because of the missing route.
```
ipsec auto --down north-east-sn
"north-east-sn": terminating SAs using this connection
"north-east-sn" #3: ESP traffic information: in=84B out=84B
```
```
ipsec auto --down north-east-gw
"north-east-gw": terminating SAs using this connection
"north-east-gw" #1: sent INFORMATIONAL request to delete IKE SA
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 0.5 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 1 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 2 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 4 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 8 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 16 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: retransmission; will wait 32 seconds for response
"north-east-gw" #1: ESTABLISHED_IKE_SA: 60 second timeout exceeded after 7 retransmits.  No response (or no acceptable response) to our IKEv2 message
"north-east-gw" #2: ESP traffic information: in=84B out=84B
```

I had several attempts to do this automatically, but without success.

e.g.
Only setting and deleting the route when ${PLUTO_PEER} equal ${PLUTO_PEER_CLIENT_NET}, because you only need the route in this case. Unfortunately this doesn't work if the peer is behind a nat router.
Another solution was to check `ip xfrm state` for active states to that peer, before deleting a route, but this could be too inaccurate.

Therefore I think the only way to fix this scenario is with different metric settings.


wofferl:
> The kernel has space for only one mark:
> 
> ```
> 	__u32			sk_mark;
> ```
> 
> are connections required to use the same mark for the same remote?

no, i think not and yes this seems to solve the problem. You only need the route for the gateway tunnel, so when in this test scenario mark-out is removed from the subnet tunnel, where it is not needed, it doesn't add/remove the table 50 route and you also can remove the metric parameters.

Seems an easy solution :)

Nevertheless, the metric handling is needed if you wanted to use it

cagney:
> Nevertheless, the metric handling is needed if you wanted to use it

agreed; tests results are clean.  Just waiting to see if @antonyantony had any thoughts.

cagney:
merged!

PR #1716, opened 2024-05-22
kernel_xfrm: record extended ack from netlink response

ueno:
This enables pluto to log any error message reported through extended ACK attributes[1] in a netlink response, to make diagnostic easier when an error occurs.  Suggested by Sabrina Dubroca.

1. https://docs.kernel.org/userspace-api/netlink/intro.html#ext-ack

### DIFF
--- a/programs/pluto/kernel_xfrm.c
+++ b/programs/pluto/kernel_xfrm.c
@@ -332,6 +332,19 @@ static void kernel_xfrm_init(struct logger *logger)
 			    "socket() in init_netlink()");
 	}
 
+	static const int on = true;
+
+	if (setsockopt(nl_send_fd, SOL_NETLINK, NETLINK_CAP_ACK,
+		       (const void *)&on, sizeof(on)) < 0) {
+		llog_errno(DEBUG_STREAM, logger, errno,
+			   "setsockopt NETLINK_CAP_ACK in kernel_xfrm_init()");
+	}
+	if (setsockopt(nl_send_fd, SOL_NETLINK, NETLINK_EXT_ACK,
+		       (const void *)&on, sizeof(on)) < 0) {
+		llog_errno(DEBUG_STREAM, logger, errno,
+			   "setsockopt NETLINK_EXT_ACK in kernel_xfrm_init()");
+	}
+
 	init_netlink_rtm_fd(logger);
 	init_netlink_xfrm_fd(logger);
 
@@ -385,6 +398,32 @@ static void kernel_xfrm_flush(struct logger *logger)
 	sendrecv_xfrm_msg(&state.n, NLMSG_ERROR, &rsp,
 			  "flush", "state",
 			  &recv_errno, logger);
+}
+
+static void llog_ext_ack(lset_t rc_flags, struct logger *logger,
+			 const struct nlmsghdr *n)
+{
+	if (n->nlmsg_type != NLMSG_ERROR ||
+	    !(n->nlmsg_flags & NLM_F_ACK_TLVS)) {
+		return;
+	}
+
+	struct nlmsgerr *err = (void *)n + NLMSG_HDRLEN;
+	size_t offset = sizeof(*err);
+	if (!(n->nlmsg_flags & NLM_F_CAPPED)) {
+		offset += err->msg.nlmsg_len - NLMSG_HDRLEN;
+	}
+
+	for (const struct nlattr *attr = nl_getattr(n, &offset);
+	     attr != NULL; attr = nl_getattr(n, &offset)) {
+		if ((attr->nla_type & NLA_TYPE_MASK) == NLMSGERR_ATTR_MSG) {
+			const char *msg = nl_getattrvalstrz(n, attr);
+			if (msg) {
+				llog(rc_flags, logger, "netlink ext_ack: %s",
+				     msg);
+			}
+		}
+	}
 }
 
 /*
@@ -505,6 +544,7 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 		if (rsp.u.e.error != 0) {
 			llog_error(logger, -rsp.u.e.error,
 				   "netlink response for %s %s", description, story);
+			llog_ext_ack(RC_LOG, logger, &rsp.n);
 			return false;
 		}
 		/*
@@ -515,6 +555,7 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 		 */
 		ldbg(logger, "%s() netlink response for %s %s included non-error error",
 		     __func__, description, story);
+		llog_ext_ack(DEBUG_STREAM, logger, &rsp.n);
 		/* ignore */
 	}
 	if (rbuf == NULL) {
--- a/include/netlink_attrib.h
+++ b/include/netlink_attrib.h
@@ -46,4 +46,8 @@ void nl_addattrstrz(struct nlmsghdr *n, int maxlen, int type,
 		const char *str);
 void nl_addattr32(struct nlmsghdr *n, int maxlen, int type, const uint32_t data);
 
+const struct nlattr *nl_getattr(const struct nlmsghdr *n, size_t *offset);
+const char *nl_getattrvalstrz(const struct nlmsghdr *n,
+			      const struct nlattr *attr);
+
 #endif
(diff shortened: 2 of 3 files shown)


ueno:
I'm currently testing this using the KVM tests (e.g., pluto/ikev2-algo-sha2-03), though haven't managed to receive ACK TLVs, even if NETLINK_{CAP,EXT}_ACKS is set with setsockopt. Perhaps the support might not be part of the kernel 6.8.9?

letoams:
On Tue, 21 May 2024, Daiki Ueno wrote:

> I'm currently testing this using the KVM tests (e.g., pluto/ikev2-algo-sha2-03), though haven't managed to receive ACK
> TLVs, even if NETLINK_{CAP,EXT}_ACKS is set with setsockopt. Perhaps the support might not be part of the kernel 6.8.9?

So I have two questions:

1) What happens when we try to enable this on kernels that do not
    support this? Will it be ignored or throw an error?
2) When was this support added to the kernel? e.g. is it safe for
    us to assume the support is there? (eg 2.x, 3.x? 6.x ?)

Paul

qsn:
6.8.9 should work for your tests, but you will only get the extack response if your netlink request is invalid and the state/policy can't be installed. This may not be easy to trigger since I expect libreswan to generate correct requests in almost all cases. I'm assuming you validate strings taken from the user's config before converting them into netlink requests, so generating a request with a bogus ICV length or non-existent algorithm ("foobar" instead of "rfc4106(gcm(aes))", or some algorithm that's not supported on the current kernel), or a bogus IP mask length, or 42 instead of XFRM_MODE_TUNNEL, etc, is likely not easy.

The kernel's IPsec implementation has been providing extacks since 6.2 (the bulk of the changes went into 6.1), but extacks existed for a long time before that (4.12, see [merge commit](https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/?id=ce07183282975026716107d36fd3f5f93de76668)).

In case the kernel just doesn't support extack, the setsockopt enabling extack on the socket should fail and do nothing (and the return value is ignored in this patch). No extack information will be returned in the kernel's answer.

In case the kernel supports extack but xfrm doesn't fill that information, the userspace application will just not get any extra details on why the request failed. This can also happen on 6.2+ kernels if the kernel does not explicitly set the error (which I think is the case for all ENOMEM type of situations, because I didn't see any value in adding a string that would basically repeat ENOMEM's meaning - and maybe a few other cases).

ueno:
I've confirmed that, if I add `req.p.mode = 42` in netlink_add_sa, I get:
```console
ERROR: "westnet-eastnet-ipv4-psk-ikev2" #2: netlink response for Add SA esp.b455
0ed8@192.1.2.23: Invalid argument (errno 22)
"westnet-eastnet-ipv4-psk-ikev2" #2: netlink ext_ack: Unsupported mode
```
from ikev2-algo-sha2-01 test. Not sure how this can be exercised properly in the test suite though.

programs/pluto/kernel_xfrm.c
```diff
@@ -387,6 +400,36 @@ static void kernel_xfrm_flush(struct logger *logger)
 			  &recv_errno, logger);
 }
 
+static void print_ext_ack(const struct nlmsghdr *n, struct jambuf *buf)
```
cagney, line 403: Can you please make the signature `jam_ext_ack(struct jambuf *, const struct nlmsghdr *)` so it's like the other jam functions.

cagney: never mind, I noticed something else

programs/pluto/kernel_xfrm.c
```diff
@@ -505,6 +548,9 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 		if (rsp.u.e.error != 0) {
 			llog_error(logger, -rsp.u.e.error,
 				   "netlink response for %s %s", description, story);
+			LLOG_JAMBUF(RC_LOG, logger, buf) {
+				print_ext_ack(&rsp.n, buf);
+			}
```
cagney, line 553: I see print_ext_ack() can return without emitting anything which could lead to blank log and debug lines.  Can I suggest changing `print_ext_ack()` to:
```
  llog_ext_ack(lset_t rc_flags, struct logger *logger, const struct nlmsghdr *n)
```
with call:
```
llog_ext_ack(RC_LOG, logger, &rsp.n);
```
and ...

programs/pluto/kernel_xfrm.c
```diff
@@ -515,6 +561,9 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 		 */
 		ldbg(logger, "%s() netlink response for %s %s included non-error error",
 		     __func__, description, story);
+		LDBGP_JAMBUF(DBG_BASE, logger, buf) {
```
cagney, line 564: ... this code changed to:
```
if (DBGP(DBG_BASE)) {
    LDBG_log(logger, "%s() netlink response for %s %s included non-error error",
		     __func__, description, story);
   llog_ext_ack(DEBUG_STREAM, logger, &rsp.n);
}

Comment (cagney) [reply programs/pluto/kernel_xfrm.c:564]: PS: you could even merge the "netlink response for ..." `into llog_ext_ack()`

File: programs/pluto/kernel_xfrm.c [hunk 4]
```diff
@@ -387,6 +400,36 @@ static void kernel_xfrm_flush(struct logger *logger)
...
+	     attr != NULL; attr = nl_getattr(n, &offset)) {
+		if ((attr->nla_type & NLA_TYPE_MASK) == NLMSGERR_ATTR_MSG) {
+			const char *msg = nl_getattrvalstrz(n, attr);
+			if (msg) {
+				if (first) {
+					jam(buf, "netlink ext_ack: %s", msg);
+					first = false;
+				} else {
+					jam(buf, "; %s", msg);
+				}
```
cagney, line 427: If it makes your life easier, you could log one message per log line.  Otherwise, you can `break` out of the `LLOG_JAMBUF()`

qsn:
> I've confirmed that, if I add `req.p.mode = 42` in netlink_add_sa, I get:
> 
> ```
> ERROR: "westnet-eastnet-ipv4-psk-ikev2" #2: netlink response for Add SA esp.b455
> 0ed8@192.1.2.23: Invalid argument (errno 22)
> "westnet-eastnet-ipv4-psk-ikev2" #2: netlink ext_ack: Unsupported mode
> ```

Nice :)

> from ikev2-algo-sha2-01 test. Not sure how this can be exercised properly in the test suite though.

Does libreswan prevent the combinations of options that the kernel doesn't support? AH and IPcomp don't work with UDP encap or TFCPAD, IPcomp only works with transport/tunnel mode (not BEET), TFCPAD only works in tunnel mode, NIC offload only works with ESP but not AH (and for crypto offload, not with UDP/TCP encap or TFCPAD). If some of those are considered valid by libreswan, the kernel should reject them and return an extack message.

letoams:
On May 24, 2024, at 09:23, qsn ***@***.***> wrote:
> 
> ﻿Does libreswan prevent the combinations of options that the kernel doesn't support? AH and IPcomp don't work with UDP encap or TFCPAD, IPcomp only works with transport/tunnel mode (not BEET), TFCPAD only works in tunnel mode, NIC offload only works with ESP but not AH (and for crypto offload, not with UDP/TCP encap or TFCPAD). If some of those are considered valid by libreswan,

Nope, libreswan won’t let you do those.
The easiest way to test this is to add an —impair that sets a bogus value.

programs/pluto/kernel_xfrm.c
ueno: Done.

> PS: you could even merge the "netlink response for ..." into llog_ext_ack()

I left it as-is, as it might make testing a little harder (some tests are expecting this line and extack might not be available if kernel is older.

programs/pluto/kernel_xfrm.c
ueno: Done.

cagney:
testing

cagney:
I'll change:
```
@@ -515,6 +555,7 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
                 */
                ldbg(logger, "%s() netlink response for %s %s included non-error error",
                     __func__, description, story);
+               llog_ext_ack(DEBUG_STREAM, logger, &rsp.n);
                /* ignore */
        }
```
to
```
  if (DBGP(DBG_LOG)) {
    LDBG_log(logger, ...)
    llog_ext_ack(DEBUG_STREAM, ...)
}
```
so it doesn't always debug log

cagney:
pushed

PR #1729, opened 2024-06-12
WIP: kernel NATT keepalive

paulwouters:


cagney:
Wow, whack doesn't have --nat-keepalive, just --no-nat-keepalive!

Suggest, passing the value to pluto as a string.  That way connections.c can decide what to do with no,yes,NNN and keep things compatible.

cagney:
For historic reasons the names needed to be different - I'm guessing that is part of the reasoning behind `keep-alive=<seconds>` and `nat-keepalive={yes,no}`,  but that limitation's mostly removed.  Should the connection just allow `keep-alive=<seconds>`

programs/pluto/linux-copy/linux/xfrm.h
```diff
@@ -321,13 +315,7 @@ enum xfrm_attr_type_t {
...
 	XFRMA_SA_DIR,		/* __u8 */
-	XFRMA_NAT_KEEPALIVE_INTERVAL,   /* __u32 in seconds for NAT keepalive */
-        XFRMA_IPTFS_DROP_TIME,  /* __u32 in: usec to wait for next seq */
-        XFRMA_IPTFS_REORDER_WINDOW, /* __u16 in: reorder window size (pkts) */
-        XFRMA_IPTFS_DONT_FRAG,  /* out: don't use fragmentation */
-        XFRMA_IPTFS_INIT_DELAY, /* __u32 out: initial packet wait delay (usec) */
-        XFRMA_IPTFS_MAX_QSIZE,  /* __u32 out: max ingress queue size (octets) */
-        XFRMA_IPTFS_PKT_SIZE,   /* __u32 out: size of outer packet, 0 for PMTU */
+	XFRMA_NAT_KEEPALIVE_INTERVAL,	/* __u32 in seconds for NAT keepalive */
 	__XFRMA_MAX
```
cagney, line 319: this is in mainline; ya!

programs/pluto/kernel.h
```diff
@@ -296,23 +295,17 @@ struct kernel_ops {
...
-	err_t (*directional_ipsec_sa_is_enabled)(struct logger *);
-	bool (*directional_ipsec_sa)(struct child_sa *child);
 	bool (*poke_ipsec_policy_hole)(int fd, const struct ip_info *afi, struct logger *logger);
-	bool (*detect_nic_offload)(const char *name, const struct logger *logger);
+	bool (*detect_nic_offload)(const char *name, struct logger *logger);
 	bool (*poke_ipsec_offload_policy_hole)(struct nic_offload *nic_offload, struct logger *logger);
-
-	/* extensions */
-	const struct kernel_ipsec_interface *ipsec_interface;
+	bool (*detect_sa_direction)(struct logger *logger);
```
cagney, line 308: There's:
```
        err_t (*directional_ipsec_sa_is_enabled)(struct logger *);
        bool (*directional_ipsec_sa)(struct child_sa *child);
```
I'm guessing the first is equivalent to `detect_sa_direction()`

lib/libipsecconf/keywords.c
```diff
@@ -0,0 +1,663 @@
...
+  { "narrowing",  kv_conn,  kt_sparse_name,  KNCF_IKEv2_ALLOW_NARROWING, &yn_option_names, NULL, },
+  { "pam-authorize",  kv_conn,  kt_sparse_name,  KNCF_PAM_AUTHORIZE, &yn_option_names, NULL, },
+  { "send-redirect",  kv_conn,  kt_sparse_name,  KNCF_SEND_REDIRECT, &yna_option_names, NULL, },
+  { "redirect-to",  kv_conn,  kt_string,  KSCF_REDIRECT_TO, NULL, NULL, },
+  { "accept-redirect",  kv_conn,  kt_sparse_name, KNCF_ACCEPT_REDIRECT, &yn_option_names, NULL, },
+  { "accept-redirect-to",  kv_conn,  kt_string, KSCF_ACCEPT_REDIRECT_TO, NULL, NULL, },
+  { "pfs",  kv_conn,  kt_sparse_name,  KNCF_PFS, &yn_option_names, NULL, },
+
+  { "nat-keepalive",  kv_conn,  kt_seconds,  KNCF_NAT_KEEPALIVE, NULL, NULL, },
+
```
cagney, line 453: Could I suggest:

- `nat-keepalive-interval` for the timer
  + mumble something about renaming global as well, since that is now possible
  + there's `retransmit-interval`, hence suffix

- `nat-keepalive-type={AUTO,kernel,user}` to determine where it is done

that way the on/off flag `nat-keepalive{yes,no}` is kept simple

programs/pluto/connections.h
```diff
@@ -274,6 +187,7 @@ struct config {
 	deltatime_t retransmit_timeout; /* max time for one packet exchange attempt */
 	uintmax_t sa_ipsec_max_bytes;
 	uintmax_t sa_ipsec_max_packets;
+	uintmax_t sa_keepalive;
```
cagney, line 190: deltatime_t might be better as the code parsing times returns that

lib/libipsecconf/keywords.c
cagney: > mumble something about renaming global as well, since that is now possible

or drop the global config option.  If this is made per-connection then encouraging `conn %default` is better?
(yes when using ipsec.conf, no when some other tool tries to send us json)

Issue #1156, opened 2023-05-30
pluto hangs in connect() when configured with Opportunistic IPsec over TCP


ueno:
With the following configuration, pluto seems to hang in connect():

```console
$ cat /etc/ipsec.d/oa.conf
conn private
        type=tunnel
        authby=null
        leftid=%null
        rightid=%null
        left=%defaultroute
        right=%opportunisticgroup
        negotiationshunt=hold
        failureshunt=drop
        ikev2=insist
        auto=route
        enable-tcp=yes
        tcp-remoteport=4500

$ ping 10.0.0.3
...
20 packets transmitted, 0 received, 100% packet loss, time 19460ms
```

It works if `enable-tcp` is commented. Here is a backtrace:
```console
#0  0x00007fce6d17548b in connect () from /lib64/libc.so.6
#1  0x0000559b40089edd in connect_to_tcp_endpoint (local_dev=0x559b41c04388, 
    remote_endpoint=..., logger=0x559b41c094b8)
    at /usr/src/debug/libreswan-4.11-1.fc38.x86_64/programs/pluto/iface_tcp.c:487
#2  0x0000559b400456ec in initiate_v2_IKE_SA_INIT_request (c=0x559b41c0b5c8, 
    predecessor=0x0, policy=25904095257, try=1, inception=0x7ffc115a01b0, 
    sec_label=..., background=true, logger=0x7ffc115a4e00)
    at /usr/src/debug/libreswan-4.11-1.fc38.x86_64/programs/pluto/ikev2_ike_sa_init.c:545
#3  0x0000559b4009ab0b in initiate_ondemand_body (b=b@entry=0x7ffc115a0b00)
    at /usr/src/debug/libreswan-4.11-1.fc38.x86_64/programs/pluto/initiate.c:1006
#4  0x0000559b4009bc2c in initiate_ondemand (
    packet=packet@entry=0x7ffc115a2d20, by_acquire=by_acquire@entry=true, 
    background=background@entry=true, sec_label=..., 
    logger=logger@entry=0x7ffc115a4e00)
    at /usr/src/debug/libreswan-4.11-1.fc38.x86_64/programs/pluto/initiate.c:1027
#5  0x0000559b40075751 in netlink_acquire (logger=0x7ffc115a4e00, 
    n=0x7ffc115a0d10)
    at /usr/src/debug/libreswan-4.11-1.fc38.x86_64/programs/pluto/kernel_xfrm.c:1874
[...]

PR #1746, opened 2024-07-02
tcp: call kernel_ops->poke_ipsec_policy_hole before connect

ueno:
This fixes ondemand initiation with TCP.  Without the policy hole, a
TCP handshake will not complete, as it cannot receive SYN-ACK packet
in plaintext and thus connect blocks until timeout.

Fixes: #1156


cagney:
cherry-picked; thanks!

PR #1756, opened 2024-07-12
Initial support for draft-ietf-ipsecme-ikev2-qr-alt-00

vukasink:
A note: While I was implementing this, the code in crypt_dh_v2.c changed with aab8bb39a8dd880128284d40001fde638e7f427f commit. I used the unified approach that was there prior to that commit  (via static void calc_skeyseed_v2) to recalculates the keys with PPK key. Then when I wanted to create PR, I saw calc_skyseed_v2 was removed. I did a rebase by making (according to the new convention) another whole function that recalculates the keys. That function contains a lot of duplicate code from function calc_v2_keymat. I think this is the fastest approach to get this PR reviewed and merged. After that is done, we can again extract the duplicate code back to static void calc_skeyseed_v2 if needed.

In addition, there are some implementation notes in the message of c7e79bbfd1d932f6ffe7b44706b5b2c133bafa6f commit.

I'll be happy to make changes to the PR if needed.


vukasink:
BTW. I have a couple of more tests that I am preparing that I will add to this or some future new PR.

cagney (changes requested):
I can see why you're renaming it.  This call:
```
		chunk_t *ppk_id;
		get_connection_ppk_initiator(ike->sa.st_connection, &ppk_id);
```
makes little sense.

Is one still only for the initiator and the other for the responder?  If so I'd suggest something like:
```
get_connection_initiator_ppk_and_ppk_id();
get_connection_responder_ppk();
```

so it is also clear which function should be called by which half of the exchange

cagney (changes requested):
Can I suggest commenting the two new parameters.  Something like:
```
		const shunk_t ppk = get_connection_ppk_and_ppk_id(ike->sa.st_connection,
								  /*find_any*/true,
								  /*index*/0,
								  &ppk_id);
```
so it is clear what each parameter is for. 

Perhaps one day we'll have `.find_any = true` :-)

lib/libswan/constants.c
```diff
@@ -2108,13 +2108,15 @@ static const char *const v2_notification_status_name[] = {
 
 static const char *const v2_notification_private_name[] = {
 #define S(E) [E - v2N_NULL_AUTH] = #E
-	S(v2N_NULL_AUTH),	/* 40960, used for mixed OE */
```
cagney, line 2111: Unfortunately this isn't so easy.  The result will be an array containing `50209 - 40960 + 1` elements, oops.
There are two ways to handle this:

- create multiple enum_name tables and chain them together
- convert this to a sparse_name table which contains name+value pairs

the choice is yours

cagney (commented):
I see the draft is up to revision 3.    I can either:

- merge as-is with tweaks
- merge just the tweaked prep patches
- something else

Any preference?

programs/pluto/state.h
```diff
@@ -571,8 +571,12 @@ struct state {
...
 	 * Post-quantum Preshared Key variables (v2)
 	 */
-	bool st_ppk_used;			/* both ends agreed on PPK ID and PPK */
-	bool st_v2_ike_ppk_enabled;		/* does remote peer support PPK? */
+	bool st_ppk_ike_auth_used;			/* both ends agreed on PPK ID and PPK (RFC 8784) */
+	enum {
+		PPK_DISABLED = 0,
+		PPK_IKE_AUTH = 1,		/* RFC 8784 */
+		PPK_IKE_INTERMEDIATE = 2,	/* draft-ietf-ipsecme-ikev2-qr-alt-00 */
+	} st_v2_ike_ppk;
```
cagney, line 579: If draft 3 contains substantive changes an option is to just factor out this change so it can be merged.
Up to you.

vukasink: I've updated the PR to qr-alt-04 version in 67796a7; no structural changes, just Notify rename

lib/libswan/constants.c
vukasink: Thanks for catching this. Fixed in 25fa57e.

vukasink:
> I can see why you're renaming it. This call:
> 
> ```
> 		chunk_t *ppk_id;
> 		get_connection_ppk_initiator(ike->sa.st_connection, &ppk_id);
> ```
> 
> makes little sense.
> 
> Is one still only for the initiator and the other for the responder? If so I'd suggest something like:
> 
> ```
> get_connection_initiator_ppk_and_ppk_id();
> get_connection_responder_ppk();
> ```
> 
> so it is also clear which function should be called by which half of the exchange

`get_connection_ppk_and_ppk_id` is used (ATM) only by initiator, while `get_connection_ppk` is used by both initiator and responder. I guess having "initiator" in the first and nothing in the second function name would be even more confusing. I suggest to leave it as it is now, but I do not have a strong opinion about it, you can choose.

vukasink:
Okay, I think I covered all your comments.

I am not sure what is the best approach regarding changes to INTERMEDIATE code, whether to apply first this PR or the other one (multiple KEs). Obviously, I have a preference towards one option, but I am fine with the other one as well, if you think it's easier approach overall.

vukasink:
My fixes weren't building on the top of current main. So I rebased locally, fixed the build errors and force-pushed. Hopefully the discussion above is still understandable with my rebase+force push.

PR #1791, opened 2024-08-30
Update ckaid.c,fix incorrectly using comma to end

yuncang123:
I use clang to build,and find the little problem.


yuncang123:
@cagney @paulwouters

cagney:
thanks!

PR #1814, opened 2024-09-23
pluto: allow nic-offload=packet to configure lifetime limits

PatrisiousHaddad:
Previously nic-offload=packet wasn't allowed to configure lifetime byte/packet limits for SA.

Simply since there was no driver/HW that supports that. That is no longer the case since Linux kernel 6.7+ which added support for lifetime limit configuration. Specifically from the following commit onwards
https://patchwork.kernel.org/project/netdevbpf/patch/20231012192750.124945-16-saeed@kernel.org/#25551683

### DIFF
--- a/programs/pluto/kernel_xfrm.c
+++ b/programs/pluto/kernel_xfrm.c
@@ -1586,19 +1586,10 @@ static bool netlink_add_sa(const struct kernel_state *sa, bool replace,
 	req.p.reqid = sa->reqid;
 	ldbg(logger, "%s() adding IPsec SA with reqid %d", __func__, sa->reqid);
 
-	if (sa->nic_offload.dev == NULL /* i.e., no offload */ ||
-	    sa->nic_offload.type != KERNEL_OFFLOAD_PACKET) {
-		req.p.lft.soft_byte_limit = sa->sa_max_soft_bytes;
-		req.p.lft.hard_byte_limit = sa->sa_ipsec_max_bytes;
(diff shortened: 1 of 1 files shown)


paulwouters:
What would happen if we try to set the lifetime limit on a kernel < 6.7 ?
Would it silently ignore it? Which would be bad in itself but not fatal
Would it fail? Which would be worse as we would break most deployments out there.

Do we need to introduce a kernel version check for this ?

PatrisiousHaddad:
**Would it fail? Sadly yes.**

For kernels older than 6.7 , SA creation would fail with EINVAL because of check we have in our driver that prevents SA creation with limit set.

**Failure after "ipsec auto --up eth2":**
"eth2" #1: IKE SA established but initiator rejected Child SA response
**From log:**
ERROR: "eth2" #8: netlink response for Add SA esp.216f71d8@2.2.2.3: Invalid argument (errno 22)
"eth2" #8: Warning: Adding IPsec SA to failed - NIC packet esp-hw-offload possibly not available for the negotiated parameters

I know that breaks backwards compatibility for old kernels, but IMO this behavior still is kind of more acceptable than what happens now , where right now for kernels < 6.7 the behavior is that the customer would request lifetime limit and it would simply ignore its request and initialize IPsec successfully without ... (though that is transparent to him if  he checks the state that limit wasnt configured).

What do you think ? does it need a kernel version check or is it acceptable without ?

paulwouters:
See https://github.com/libreswan/libreswan/pull/1844

Issue #1812, opened 2024-09-22
accepting ipsec-interface with conflicting physical device


cagney:
In `ikev1-xfrmi-04` and `ikev2-xfrmi-04` on east, two connections are loaded:
```
conn northnet-eastnet
	right=192.0.2.254  # eth1
	ipsec-interface=yes
conn westnet-eastnet
        right=192.1.2.23 # eth0
	ipsec-interface=yes
```
depending on which connection installs the ipsec-interface, the result is either:
```
-X: ipsec1@eth1: <NOARP,UP,LOWER_UP> mtu 1500 state UNKNOWN
+X: ipsec1@eth0: <NOARP,UP,LOWER_UP> mtu 1500 state UNKNOWN
```
is this valid, or should the second connection that attempts to add an interface reject things?

Issue #1807, opened 2024-09-19
create kernel ipsec-interface (pseudo device) during orient? (but not add addresses)


cagney:
This follows up

- [ ] #1795

@bleve suggested, off line, creating the kernel ipsec-interface when the connection is being added except that doesn't work because things aren't yet oriented.

So the next alternative, clearly :-), is to create the kernel ipsec-interface during orient (and presumably toss it when disorienting).
This would also suggest a split between creating an kernel ipsec-interface and populating it with addresses (the latter possibly delayed until in `route.[hc]`, but it depends on `interface-ip=` and/or` sourceip=`).

@wofferl @antonyantony thoughts

cagney:
somewhat supporting this idea, the test `ipsec-interface-15-blocked-ikev2-xfrmi` fails with:
```
"westnet6-eastnet6": added IKEv2 connection
"westnet4-eastnet4" #5: initiating IKEv2 connection to 192.1.2.23 using UDP
"westnet4-eastnet4" #5: sent IKE_SA_INIT request to 192.1.2.23:UDP/500
"westnet4-eastnet4" #5: processed IKE_SA_INIT response from 192.1.2.23:UDP/500 {cipher=AES_GCM_16_256 integ=n/a prf=HMAC_SHA2_512 group=DH19}, initiating IKE_AUTH
"westnet4-eastnet4" #5: sent IKE_AUTH request to 192.1.2.23:UDP/500
"westnet4-eastnet4" #5: initiator established IKE SA; authenticated peer using authby=secret and ID_FQDN '@east'
ERROR: "westnet4-eastnet4" #6: ipsec-interface ipsec1 with ID 1 exists but is not a valid: IFLA_XFRM_LINK attribute is zero, check 'ip -d link show dev ipsec1'
"westnet4-eastnet4" #6: CHILD SA failed: TEMPORARY_FAILURE
"westnet4-eastnet4" #5: IKE SA established but initiator rejected Child SA response
"westnet4-eastnet4" #6: sent INFORMATIONAL request to delete larval Child SA using IKE SA #5
ERROR: "westnet4-eastnet4" #6: ipsec-interface ipsec1 with ID 1 exists but is not a valid: IFLA_XFRM_LINK attribute is zero, check 'ip -d link show dev ipsec1'
EXPECTATION FAILED: "westnet4-eastnet4" #6: (*child) == ((void *)0) (teardown_child() +1238 programs/pluto/routing.c)
```
[...]

PR #1815, opened 2024-09-23
Kernel/orient ipsec interface

cagney:
Move the code looking for and, when needed, creating the Kernel's ipsec-interface to orient().  Should the operation fail then leave the connection unoriented.  Conversely, disorient() removes the interface.  The routing code is still responsible for adding and removing addresses.

This way the routing code doesn't have to recover from a completely broken ipsec-interface.

The tests `ipsec-interface-15-blocked-ikev2-xfrmi` and `ipsec-interface-15-order-ikev2-xfrmi` are updated to try various sequences of:

- `ip route add`
- `ip addr add`
- `ipsec add` (which triggers orient)
- `ipsec up`
- `ipsec down`

for instance:

```
west #
 ipsec add westnet4-eastnet4
ERROR: "westnet4-eastnet4": existing ipsec-interface ipsec1 is not valid: IFLA_XFRM_LINK attribute is zero, check 'ip -d link show dev ipsec1'
"westnet4-eastnet4": added unoriented IKEv2 connection (neither left=192.1.2.45 nor right=192.1.2.23 match an interface)
```

This addresses:

- [x] #1807 
- [ ] #1812 things are more deteministic

@wofferl @antonyantony thoughts?

### DIFF
--- a/programs/pluto/ipsec_interface.c
+++ b/programs/pluto/ipsec_interface.c
@@ -667,31 +667,88 @@ diag_t parse_ipsec_interface(struct config *config,
 	return NULL;
 }
 
-void add_ipsec_interface(struct connection *c)
+bool add_ipsec_interface(struct connection *c,
+			 const struct iface_device *iface)
 {
 	ipsec_interface_buf ifb;
-	VERBOSE_DBGP(DBG_BASE, c->logger, "adding %s to connection",
-		str_ipsec_interface_id(c->config->ipsec_interface.id, &ifb));
+	VERBOSE_DBGP(DBG_BASE, c->logger, "adding %s@%s to connection",
+		     str_ipsec_interface_id(c->config->ipsec_interface.id, &ifb),
+		     iface->real_device_name);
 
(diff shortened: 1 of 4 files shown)


cagney:
Hmm, this approach doesn't work with NetBSD because that OS expects the ipec-interface to specify both ends of the tunnel.

cagney:
> Hmm, this approach doesn't work with NetBSD because that OS expects the ipec-interface to specify both ends of the tunnel.

NetBSD, it seems has bigger problems with its ipsec-interface code.  For instance:

- it generates an unbound kernel policy that happens to match the physical interface, but new school is to have the kernel policy match and be bound to the ipsec interface (probably with ::/0 et.al)
- the msgid needed to bind the kernel state to the generated policy isn't available

so it can be ignored for now.

As an side, on first blush OpenBSD seems to be close to Linux (probably because the code was only added this year).  And FreeBSD sits somewhere in the middle (it generates a kernel policy bound to the ipsec interface).

cagney:
Off list discussion's concluded this should be merged.

PR #1830, opened 2024-10-02
ikev2: add support for multiple key exchanges

ueno:
This is my attempt to add support for multiple key exchanges as defined in RFC 9370. As the standard allows 7 additional key exchanges at maximum, this patch extends the ike= configuration line to include them in a form:
<encr>-<prf>-<DH>-<addke1>-<addke2>- ... -<addke7>.

Note: Rekeying is still not supported


cagney:
I'm not sure about the syntax.  I'm guessing it is now:
```
ike=ENCR[-PRF][-INTEG]{-DH}*
```
for instance:
```
ike=aes-sha2-sha1-dh19-dh20-pq5-dh1
```
if someone were to later add a new transform 13 (definitely unlucky) we'd be in strife over how to differentiate between it and these other algorithms.

This leads me to consider alternate syntaxes like:
```
ike=aes-sha2-sha1-dh19;addke=dh20-pq5-dh1
```
have you any thoughts on that.

Also, how does this work with @vukasink 

- [ ] #1756 from @vukasink 

my hunch is that `pq5` above would identify that the ppk code, and not the dh code would be used for that round?  Or are the two orthogonal?

ueno:
> if someone were to later add a new transform 13 (definitely unlucky) we'd be in strife over how to differentiate between it and these other algorithms.
> 
> This leads me to consider alternate syntaxes like:
> 
> ```
> ike=aes-sha2-sha1-dh19;addke=dh20-pq5-dh1
> ```
> 
> have you any thoughts on that.

That sounds like a good idea; I'll try to implement it.

> Also, how does this work with @vukasink
> 
>     * [ ]  [Initial support for draft-ietf-ipsecme-ikev2-qr-alt-00 #1756](https://github.com/libreswan/libreswan/pull/1756) from @vukasink
> 
> 
> my hunch is that `pq5` above would identify that the ppk code, and not the dh code would be used for that round? Or are the two orthogonal?

To my understanding, those concepts are orthogonal; the new ADDKE[1-7] transform types take the ID from DH groups, while a PPK is not classified as DH. On the other hand, draft-kampanakis-ml-kem-ikev2 defines new transform types for ML-KEM based key exchanges, which would be specified in the same addke configuration syntax. I have an early [wip](https://github.com/ueno/libreswan/tree/wip/mlkem) for that.

vukasink:
ikev2-qr-alt draft seems to be clear on interaction with RFC 9370 (i.e., there is none): 

> Once the PPK is negotiated in the last IKE_INTERMEDIATE exchange, the IKE SA keys are recalculated. Note that if the IKE SA keys are also recalculated as the result of the other actions performed in the IKE_INTERMEDIATE exchange (for example, as defined in [[RFC9370](https://www.ietf.org/archive/id/draft-ietf-ipsecme-ikev2-qr-alt-00.html#RFC9370)]), then applying PPK MUST be done after all of them, so that recalculating IKE SA keys with PPK is the last action before they are used in the IKE_AUTH exchange.

But I am not familiar with RFC 9370 so not totally sure there really is no interaction.

cagney:
Ah.  The number of IKE_INTERMEDIATE exchanges is pre-negotiated.   That means that if this patch were to be applied, the PPK code would need to be changed so it is called near the end of process_v2_IKE_INTERMEDIATE_request/response the last IKE_INTERMEDIATE exchange is being processed.

Looking at the code, though, I see .ke_index tracking the current IKE_INTERMEDIATE exchange, but I don't see an easy way to detect it is the final exchange.  Perhaps, an additional field with the number of exchanges expected so it can be compared?

cagney:
@ueno could you rebase?

The algorithm parser is tested using algparse.  Can you add tests to that.  It is easy to run standalone, vis:
```
$ ./OBJ.linux.x86_64/programs/algparse/algparse ike=aes_gcm-sha2-dh21
algparse -v2 'ike=aes_gcm-sha2-dh21' (expect ERROR)
	AES_GCM_16-HMAC_SHA2_256-DH21
$ ./OBJ.linux.x86_64/programs/algparse/algparse -tp
algparse -v2 'esp' (expect SUCCESS)
	AES_GCM_16_256-NONE
	AES_GCM_16_128-NONE
	CHACHA20_POLY1305-NONE
	AES_CBC_256-HMAC_SHA2_512_256+HMAC_SHA2_256_128
	AES_CBC_128-HMAC_SHA2_512_256+HMAC_SHA2_256_128
algparse -v2 'esp=' (expect ERROR)
	ERROR: ESP proposal is empty
algparse -v2 'esp=aes' (expect SUCCESS)
	AES_CBC-HMAC_SHA2_512_256+HMAC_SHA2_256_128
algparse -v2 'esp=aes;modp2048' (expect SUCCESS)
ipsec algparse: ignoring ESP DH algorithm MODP2048 as PFS policy is disabled
	AES_CBC-HMAC_SHA2_512_256+HMAC_SHA2_256_128
...
```
(hmm, expect ERROR shouldn't appear on the first command)

For syntax, my current thought is to have a long term goal of always parsing some sort of canonical form such as:
```
[<1-transforms>]-[<2-transforms>]-[<3-transforms>]-[<4-transforms>]-...
```
i.e.,:
```
[<encr>]-[<prf>]-[<integ>]-[<dh>]-[<esn>]-[<addke1>]-...
```
with smarts to recognize a field is being skipped.  Currently this is somewhat true (there's magic around the order of prf-integ vs integ-prf, and esn isn't recognized).

This form can then be used (possibly with impairment enabled) to feed garbage into the proposal code where it will send garbage to the poor suffering peer under test.

Looking at the `v2_proposals.c` changes I think they are going in that direction, but additions to `algparse` would confirm that.

cagney:
time for a rebase

cagney (changes requested):
needs a full rebase, then I can kick off testing

ueno:
Yes, I'm on it.

cagney:
can pick this up post 5.3

cagney (commented):
I don't understand why this change is needed.

A simple responder, when processing IKE_SA_INIT, will:

- generate responder_KE
- compute initiator_KE+responder_KE to form the shared secret aka skeyseed
- send back IKE_SA_INIT response with responder_KE

however, since the output from the middle step - compute the shared secret - isn't needed until there's a packet to decrypt, it can be delayed, reducing the IKE_SA_INIT responder code to:

- generate responder_DH
- send back IKE_SA_INIT response

The shared secret calculation is then started once there's an encrypted IKE_AUTH or IKE_INTERMEDIATE packet (or fragment) to decrypt.

Since it's common for the server to only receive a single IKE_SA_INIT (e.g., mobile phone, or attacker) there's well established benefit in doing this - faster IKE_SA_INIT turnaround and fewer wasted CPU cycles.  However:

- the code handling untrusted packets is far more complex
- it doesn't make things faster
it takes just as long to establish the Child SA

With this in mind, I don't see any benefit in doing this with IKE_INTERMEDIATE.

- we're mixing untrusted and trusted code paths
- the peer is clearly being responsive so there's no reason to hold off on DH
- delaying the DH until the next packet doesn't make things faster

What might be possible long term, and assuming we see evidence of a need, is to compute the IKE_INTERMEDIATE DH in the background.  That way it may be ready in time for the next packet, which can speed up the exchange.

cagney (changes requested):
> ikev2: add support for multiple key exchanges

Breaking this down:

- `ietf_constants.h` + `constants.c`

As a separate patch, it can go straight in.  New additions to ietf_constants.h are a no-brainer.

There's a catch though.  It will cause check-*-enums test to fail - the output needs updating.

- proposals.h et.al.

Again this can be separated out.  It should also be accompanied to updates to algparse -t so we confirm the new behaviour.

- programs/pluto

Since DH isn't going to be delayed (for now) this will need to be reworked.

cagney:
The include and lib changes - sans the need to fix test regressions - dropped in.

Unfortunately, per #2367 - new interface required - and https://github.com/libreswan/libreswan/pull/1830#pullrequestreview-3011895549) the changes to pluto proper won't be so easy.

ueno:
@cagney thank you for the bump, and sorry for not being able to provide update on this in the last quarter.

> > ikev2: handle more cases with intermediate in request_no_skeyseed
> 
> I don't understand why this change is needed.
[...]

Yes, I remember the large part of this became unnecessary since draft-ietf-ipsecme-ikev2-qr-alt was implemented.

cagney:
@ueno btw, if you watch mainline you can see I keep merging your code!   I'm about to merge in the ike_sa_init change.  (https://github.com/cagney/libreswan/tree/multiple-ke has a rebase).

programs/pluto/ikev2_proposals.c
```diff
@@ -1745,6 +1756,47 @@ bool ikev2_proposal_to_trans_attrs(const struct ikev2_proposal *proposal,
...
+				 * a map (additional_ke_map) to ensure that.
+				 */
+				unsigned n = type - IKEv2_TRANS_TYPE_ADDKE1;
+				if (additional_ke_map[n]) {
+					passert(additional_ke_map[n]->type == type);
+					additional_ke_map[n]->group = group;
+				} else {
+					ta.ta_addke[additional_ke_index].type = type;
+					ta.ta_addke[additional_ke_index].group = group;
+					additional_ke_map[n] = &ta.ta_addke[additional_ke_index];
```
cagney, line 1796: Note to self: `additional_ke_index` should be incremented.

programs/pluto/ikev2_proposals.c
```diff
@@ -867,7 +867,9 @@ static int ikev2_process_proposals(struct pbs_in *sa_payload,
 					 */
 					if ((type == IKEv2_TRANS_TYPE_INTEG &&
 					     sentinel_transform->id == IKEv2_INTEG_NONE) ||
-					    (type == IKEv2_TRANS_TYPE_DH &&
+					    ((type == IKEv2_TRANS_TYPE_DH ||
+					      (type >= IKEv2_TRANS_TYPE_ADDKE1 &&
+					       type <= IKEv2_TRANS_TYPE_ADDKE7)) &&
```
cagney, line 872: Note to self: writing these long form is clearer as in:
```
					if ((type == IKEv2_TRANS_TYPE_INTEG &&
					     sentinel_transform->id == IKEv2_INTEG_NONE) ||
					    (type == IKEv2_TRANS_TYPE_KE &&
					     sentinel_transform->id == OAKLEY_GROUP_NONE) ||
					    (IKEv2_TRANS_TYPE_ADDKE1 <= type && type <= IKEv2_TRANS_TYPE_ADDKE7 &&
					     sentinel_transform->id == OAKLEY_GROUP_NONE)) {
						optional_transform_types |= LELEM(type);
```

PR #1844, opened 2024-10-11
linux: packet offload counters supported in 6.7+

paulwouters:


programs/pluto/kernel.h
```diff
@@ -80,6 +80,12 @@ struct nic_offload {
 	enum kernel_offload_type type;
 };
 
+struct kernel_info {
+	char *name; /* eg "Linux" */
```
cagney, line 84: Can I suggest having an enum for the OS, might as well do the streq(..., "Linux") once.

letoams:
On Fri, 11 Oct 2024, cagney wrote:

> Can I suggest having an enum for the OS, might as well do the streq(..., "Linux") once.

Sure. Since I couldn't get the BSDs to install on my test harnass, do
you have the exact strings for me?

Paul

cagney:
FreeBSD
NetBSD
OpenBSD

cagney:
Although, to be honest, until the BSDs have a need, a value for LINUX and UNKNOWN is all that's needed

PatrisiousHaddad:
Thanks Paul, this patch works fine for us.

Just confirming, this version check is actually in order to simply provide an error message from libreswan:
The behavior would still be kernels >= 6.7 it would work (I tested both lower and higher)
and kernels < 6.7 would simply return the error "Linux kernel 6.7+ required for byte/packet counters and hardware offload" 
when adding a connection ?

paulwouters:
merged slightly differently in https://github.com/libreswan/libreswan/commit/fe4253e7fc3a6f8ba285a745fc9ffcc73e7aec22#diff-4db91d65147e25bed48e091e6be59759af0a63f033be75d9a7b4c7c175ed1f76R55

PR #1987, opened 2025-01-02
addconn: improve performance of 'ipsec add' at scale

igsilya:
Issue: #2002 

Scenario:
1. We have an `ipsec.conf` with a 1000 connections specified.
2. Connections have `rightprotoport=udp/6081` or `leftprotoport=udp/6081`
3. We add each of these connections with `ipsec add <conn>` one by one, since there are other connections we don't want to disrupt in the process.

Result:
- It takes about 30 minutes to complete the configuration.

Cause:
- Every `ipsec add <conn>` results in full parsing of the large `ipsec.conf`.
- While parsing `*protoopt=udp/6081` of every connection, a very heavy `getservbyname()` function is called (takes 1.8 ms per call).

This PR attempts to solve the problem by:
1. Only loading connections that are specified on the command line (as possible), to avoid unnecessary memory allocations and extra parsing of non-interesting connections.
2. Reordering the `protoport` parsing logic to first check if the value is a number to avoid the costly `getservbyname()` call.

Results:
1. 1000x `ipsec add <conn>`:
```
        protoport      main       Patch #1    Patch #2      Both     SpeedUp
        --------------------------------------------------------------------
        udp/6081      29.6 min    24.5 sec    54.4 sec    22.5 sec     79x
        udp/geneve    22.9 min    23.9 sec    22.9 min    23.9 sec     57x
```
2. `addconn --checkconfig` (only affected by the change in `ttoport`):
```
        protoport      Before     After     SpeedUp
        -------------------------------------------
        udp/6081      1.77 sec   0.04 sec     44x
        udp/geneve    1.37 sec   1.37 sec     --
```

Technically, either one of the changes solves the problem for the main use-case with Open vSwitch (`udp/6081`), but it seemed better to fix the issue for both.

Note: I'm not particularly happy with parsing of aliases and recursive checks for `also`, but I didn't find a better way to filter out connections without breaking the logic for aliases.

Testing done with the following simple script (adjust `N` to get +/-250 connections):
```
# cat pluto-test.sh 
kill $(cat /tmp/pluto-test/pluto.pid 2>/dev/null) 2>/dev/null
sleep 2
ip netns del pluto-ns 2>/dev/null
ip netns add pluto-ns
ip -netns pluto-ns link set dev lo up
ip link del pluto-p1 2>/dev/null
ip link add pluto-p1 type veth peer name pluto-p2
ip link set pluto-p2 netns pluto-ns
ip -netns pluto-ns link set dev pluto-p2 up
ip -netns pluto-ns addr add 30.30.1.1/16 dev pluto-p2
ip -netns pluto-ns route add default via 30.30.100.254

rm -rf /tmp/pluto-test
mkdir -p /tmp/pluto-test/ipsec.d
echo '
config setup
    uniqueids=yes

conn %default
    keyingtries=%forever
    type=transport
    auto=route
    ike=aes_gcm256-sha2_256
    esp=aes_gcm256
    ikev2=insist
' > /tmp/pluto-test/ipsec.conf

touch /tmp/pluto-test/secrets
ipsec initnss --nssdir /tmp/pluto-test/ipsec.d
ip netns exec pluto-ns ipsec pluto --config /tmp/pluto-test/ipsec.conf \
      --ipsecdir /tmp/pluto-test --nssdir /tmp/pluto-test/ipsec.d \
      --logfile /tmp/pluto-test/pluto.log \
      --secretsfile /tmp/pluto-test/secrets \
      --rundir /tmp/pluto-test

sleep 2

set -o errexit

N=5
M=251
for i in $(seq 2 ${N}); do
    for j in $(seq 2 ${M}); do
        echo "
conn tunnel-${i}-${j}
    left=30.30.1.1
    right=30.30.${i}.${j}
    authby=secret
    leftprotoport=udp
    rightprotoport=udp/6081
" >> /tmp/pluto-test/ipsec.conf
        echo "
%any 30.30.${i}.${j} : PSK \"swordfish\"
" >> /tmp/pluto-test/secrets
    done
done

ip netns exec pluto-ns ipsec \
    --ctlsocket /tmp/pluto-test/pluto.ctl \
    --config /tmp/pluto-test/ipsec.conf   \
    rereadsecrets >/dev/null 2>&1

start_time_total=$(date +%s.%3N)
k=1
for i in $(seq 2 ${N}); do
    for j in $(seq 2 ${M}); do
        start_time=$(date +%s.%3N)
        ip netns exec pluto-ns ipsec \
            --ctlsocket /tmp/pluto-test/pluto.ctl \
            --config /tmp/pluto-test/ipsec.conf   \
            add tunnel-${i}-${j} \
            >> /tmp/pluto-test/add.log 2>&1
        end_time=$(date +%s.%3N)
        add_s=$(echo "scale=3; ${end_time} - ${start_time}" | bc)
        printf "%-4d %5.3f\n" ${k} ${add_s}
        let k=$(($k + 1))
    done
done
end_time_total=$(date +%s.%3N)
total_s=$(echo "scale=3; ${end_time_total} - ${start_time_total}" | bc)
printf "Total: %5.3f seconds\n" ${total_s}

kill $(cat /tmp/pluto-test/pluto.pid)
sleep 2
rm -rf /tmp/pluto-test
ip link del pluto-p1
ip netns del pluto-ns
```

### DIFF
--- a/include/ipsecconf/confread.h
+++ b/include/ipsecconf/confread.h
@@ -71,7 +71,6 @@ struct starter_end {
 	ip_address addr;
 	ip_address nexthop;
 	ip_cidr vti_ip;
-	ip_protoport protoport;
 
 	keyword_values values;
 };
--- a/include/ipsecconf/confread.h
+++ b/include/ipsecconf/confread.h
@@ -123,6 +123,7 @@ extern void parser_freeany_config_parsed(struct config_parsed **cfg);
 
 extern struct starter_config *confread_load(const char *file,
 					    bool setuponly,
+					    size_t n_allowed, char **allowed,
 					    struct logger *logger);
 
 extern void confread_free(struct starter_config *cfg);
--- a/lib/libipsecconf/confread.c
+++ b/lib/libipsecconf/confread.c
@@ -370,16 +370,6 @@ static bool validate_end(struct starter_conn *conn_st,
 		}
 	}
 
-	/* copy certificate path name */
-
-	if (end->values[KSCF_PROTOPORT].set) {
-		char *value = end->values[KSCF_PROTOPORT].string;
-		err_t ugh = ttoprotoport(value, &end->protoport);
-		if (ugh != NULL)
-			ERR_FOUND("bad %sprotoport=%s [%s]", leftright, value,
-				  ugh);
-	}
-
 	return err;
 #  undef ERR_FOUND
 }
--- a/lib/libipsecconf/confread.c
+++ b/lib/libipsecconf/confread.c
@@ -935,8 +935,76 @@ static bool init_load_conn(struct starter_config *cfg,
 	return connerr;
 }
 
+static bool should_load_conn(const struct section_list *sconn,
+			     const struct starter_conn *default_conn,
+			     const struct config_parsed *cfgp,
+			     size_t n_allowed, char **allowed)
+
+{
+	/* If the list is not provided, load everything. */
+	if (!n_allowed || !allowed)
+		return true;
+
+	/* Check if the name is in the list of allowed connections. */
+	for (size_t i = 0; i < n_allowed; i++) {
+		if (streq(allowed[i], sconn->name))
+			return true;
+	}
+
+	/*
+	 * Name was not found, check the aliases.
+	 *
+	 * First checking aliases of the default connection as they will match
+	 * for every other connection.
+	 */
+	if (default_conn->values[KSCF_CONNALIAS].set) {
+		const char *alias = default_conn->values[KSCF_CONNALIAS].string;
+
+		for (size_t i = 0; i < n_allowed; i++) {
+			if (lsw_alias_cmp(allowed[i], alias))
+				return true;
+		}
+	}
+	/*
+	 * Now checking aliases set for the connection itself and recursively
+	 * for connections referenced via 'also'.
+	 */
+	for (const struct kw_list *kw = sconn->kw; kw != NULL; kw = kw->next) {
+		if (kw->keyword.keydef->field == KSCF_ALSO) {
+			const struct section_list *addin;
+			const char *seeking = kw->string;
+
+			for (addin = TAILQ_FIRST(&cfgp->sections);
+			     addin != NULL && !streq(seeking, addin->name);
+			     addin = TAILQ_NEXT(addin, link))
+				;
+			if (addin == NULL) {
+				/*
+				 * The configuration is not correct, let it
+				 * be parsed and the error logged from the
+				 * connection loading code.
+				 */
+				return true;
+			}
+			if (should_load_conn(addin, default_conn, cfgp,
+					     n_allowed, allowed)) {
+				return true;
+			}
+		}
+
+		if (kw->keyword.keydef->field == KSCF_CONNALIAS) {
+			for (size_t i = 0; i < n_allowed; i++) {
+				if (lsw_alias_cmp(allowed[i], kw->string))
+					return true;
+			}
+		}
+	}
+	return false;
+}
+
 struct starter_config *confread_load(const char *file,
 				     bool setuponly,
+				     size_t n_allowed, char **allowed,
 				     struct logger *logger)
 {
 	/**
@@ -990,7 +1058,9 @@ struct starter_config *confread_load(const char *file,
 		 */
 		for (struct section_list *sconn = TAILQ_FIRST(&cfgp->sections);
 		     sconn != NULL; sconn = TAILQ_NEXT(sconn, link)) {
-			if (!streq(sconn->name, "%default"))
+			if (!streq(sconn->name, "%default")
+			    && should_load_conn(sconn, &cfg->conn_default, cfgp,
+						n_allowed, allowed))
 				err |= init_load_conn(cfg, cfgp, sconn,
 						      false/*default conn*/,
 						      logger);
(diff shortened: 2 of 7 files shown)


cagney (commented):
What about dropping `struct starter_end .protoport` and leaving the value as a string?  Then `set_whack_end()` can do the conversion.  Yes this means that a bogus protoport value is ignored for all but the connection being loaded, but then again, that may be a good thing.

BTW, the general trend is to leave things as strings and then let pluto do the parsing.  For instance, pluto parses `subnet=1.2.3.4/32/tcp/80`.  It means:
- addconn and whack are more consistent
- hopefully prepares some of the ground work for having pluto parse some other format

igsilya:
> What about dropping `struct starter_end .protoport` and leaving the value as a string? Then `set_whack_end()` can do the conversion.

Wouldn't that break `addconn --checkconfig` functionality, i.e. it will not fully validate the config anymore?  Being able to validate the config without communicating with `pluto` also seem like a good thing.

cagney:
> Wouldn't that break `addconn --checkconfig` functionality, i.e. it will not fully validate the config anymore?

That ship has long since departed.

Both `whack` and `addconn` leave most things to `pluto`.  That way there's only one block of code cross checking the config and applying defaults.  It also means that pluto can log what was attempted and rejected.  This is a big improvement on 4.x, where `whack`, `addconn` and `pluto` each implemented a haphazard subset of contradictory checks and defaults that had a habit of failing silently.

> Being able to validate the config without communicating with `pluto` also seem like a good thing.

True.  The file is still parsed so the basic syntax is still validated (it isn't the parser that is slow, its the verification that comes later).

One thought would be for  `addconn` and `whack` to send connections to `pluto` for additional verification (but with being adding).

Another would be to re-organize `addconn`, so that, when check-config, it goes through the motions of building each whack message but not send it.  It might then be possible to move some of the simpler checks performed by `pluto` to a library where they can be shared.

igsilya:
> > Wouldn't that break `addconn --checkconfig` functionality, i.e. it will not fully validate the config anymore?
> 
> That ship has long since departed.

OK.  So I tried the suggested solution.  Results are as follows:

Patches:
1. [addconn: do not load connections that were not requested](https://github.com/libreswan/libreswan/pull/1987/commits/5ae40dc536f4a5b59165914dc18a70e7069ab824)
2. [ttoport: check numbers before the service name lookup](https://github.com/libreswan/libreswan/pull/1987/commits/8c14ad9e3c7eb9e07444d619538dbfe847713597)
3. ipsecconf: move protoport parsing to set_whack_end (NEW)

---

1. 1000x `ipsec add`
```
    protoport      main       Patch #1    Patch #2    Patch #3      1+2       2+3       1+2+3
    -------------------------------------------------------------------------------------------
    udp/6081      29.6 min    24.5 sec    54.4 sec    55.0 sec    22.5 sec   52.0 sec   22.4 sec
    udp/geneve    22.9 min    23.9 sec    22.9 min    55.0 sec    23.9 sec   54.0 sec   23.9 sec
```
2. `addconn --checkconfig`
```
    protoport      main       Patch #1    Patch #2    Patch #3      1+2       2+3       1+2+3
    --------------------------------------------------------------------------------------------
    udp/6081      1.77 sec    1.77 sec    0.04 sec    0.04 sec    0.04 sec   0.04 sec   0.04 sec
    udp/geneve    1.37 sec    1.37 sec    1.37 sec    0.04 sec    1.37 sec   0.04 sec   0.04 sec
```

It seems like not loading connections (patch 1) is still about 2x faster than `set_whack_end()` approach (patch 3) in the `ipsec add` case and the `set_whack_end()` approach wins the `checkconfig` contest.   Having all 3 patches seems benefitial.  However, we could drop some of them.  For example, if we keep only the  `set_whack_end()` patch, then all `ipsec add` calls can be done in under a minute, which is good, though not as good as 25-ish seconds, obviously.
I'd keep patch 2 (the `ttoport` reorder) regardless, since it seem to make sense on its own even if not very beneficial in a grand scheme of things if the other changes get accepted.

What do you think?

cagney:
> 1. [addconn: do not load connections that were not requested](https://github.com/libreswan/libreswan/pull/1987/commits/5ae40dc536f4a5b59165914dc18a70e7069ab824)

See below ...

> 2. [ttoport: check numbers before the service name lookup](https://github.com/libreswan/libreswan/pull/1987/commits/8c14ad9e3c7eb9e07444d619538dbfe847713597)

I've merged this.

>  3. ipsecconf: move protoport parsing to set_whack_end (NEW)

Can you push this.  It should be good enough.

Presumably the discrepancy between 1. and 3. is all the calls to `validate_end()` expanding connections.   I suspect the work can be moved closer to the `starter_whack_add_conn()` call so it only happens when needed.  That should have the same effect as patch 1., but without the hoops.

igsilya:
> I've merged this.

Thanks!

> > 3. ipsecconf: move protoport parsing to set_whack_end (NEW)
>
> Can you push this. It should be good enough.

I rebased the PR branch and added this patch: https://github.com/libreswan/libreswan/pull/1987/commits/cef92a7be691a9d948594274924cafea0550909f

> Presumably the discrepancy between 1. and 3. is all the calls to `validate_end()` expanding connections. I suspect the work can be moved closer to the `starter_whack_add_conn()` call so it only happens when needed. That should have the same effect as patch 1., but without the hoops.

Ack.  I'll try to do some profiling around this later.  But, I agree, that having patches 2 and 3 should be good enough for now.

cagney:
merged, thanks!

PR #2024, opened 2025-01-30
testing: make jambufcheck compile with GCC 15

ueno:
Previously, when compiled with GCC 15, it fails with:
```
  libreswan-5.1/testing/programs/jambufcheck/jambufcheck.c:36:34: error: initializer-string for array of ‘char’ is too long [-Werror=unterminated-string-initialization]
     36 |                 char array[12] = "abcdefghijkl";
        |                                  ^~~~~~~~~~~~~~
```
The relevant code really needs to initialize `array` with 12 elements (10 characters + NUL + SENTINEL), but with a "..." initializer it a NUL character is automatically appended at the end. This patch switches to using a {...} to initialize the array.


cagney:
nice! merged

ueno:
Thank you, closing as it has already been merged.

PR #2049, opened 2025-02-27
Simple orthography fixes

dkg:
These were identifed by codespell, like in #1646

Again, I recommend setting up some sort of CI that will automatically catch common spelling errors.  there's no reason any human needs to automatically know how to spell these words, or to clean them up manually.  But having them misspelled gives the wrong impression about the quality of Libreswan.


cagney:
Given most of the typos are mine I agree we need to do something.
My concern is how it can be done without ending up with yet another dashboard that everyone ignores.

Issue #2088, opened 2025-03-14
pass `systemd.hostname=east` to kernel


cagney:
### What needs to be done

Suggestion from @antonyantony 

for instance:

```
     <type arch='x86_64'>hvm</type>
     <kernel>@@POOLDIR@@/@@PREFIX@@linux-transmogrify.vmlinuz</kernel>
     <initrd>@@POOLDIR@@/@@PREFIX@@linux-transmogrify.initramfs</initrd>
-    <cmdline>root=/dev/vda2 ro quiet console=tty0 plymouth.enable=0 mitigations=off biosdevname=0 net.ifnames=0 console=ttyS0,115200</cmdline>
+    <cmdline>root=/dev/vda2 ro quiet console=tty0 plymouth.enable=0 mitigations=off biosdevname=0 net.ifnames=0 console=ttyS0,115200 systemd.hostname=east</cmdline>
   </os>
```

complication is that the domains e, w, rise, set don't use this boot mechanism.

cagney:
works for east:
```
$ ./kvm sh east
Connecting to m.east
m.east 0.00: spawning: sudo virsh --connect qemu:///system console --force m.east; waiting 20 seconds
m.east 0.00: EOF opening console: b'error: The domain is not running\r\n\r\n'
m.east 0.02: spawning: sudo virsh --connect qemu:///system start m.east --console; waiting 90 seconds
m.east 6.04: boot successful (boot attempt 1 and 6.3 seconds)
m.east 6.05: got b'Password:\\s?$' prompt after 6.4 seconds; sending 'b'swan'' and waiting 10 seconds for shell prompt
m.east 6.06: got 'Last login' after 6.5 seconds; ignoring
m.east 6.07: we're in (after 6.6 seconds)!

kvmsh m.east 6.08: Escape character is ^]

[root@east ~]# cat /proc/
Display all 155 possibilities? (y or n)
[root@east ~]# cat /proc/cmdline 
root=/dev/vda2 ro quiet console=tty0 plymouth.enable=0 mitigations=off biosdevname=0 net.ifnames=0 console=ttyS0,115200 systemd.hostname=east
[root@east ~]# systemctl status hosnamer
Unit hosnamer.service could not be found.
```

cagney:
The problems are

- debian (debian{e,w,rise,set}
- fedora (fedora{e,w,rise,set}

which still use a more traditional boot (i.e., grub). Just getting {east,west,north,road} to direct boot the kernel got push back.

[...]

PR #2090, opened 2025-03-15
testing fedora vms systemd hostname in libvirt xml

antonyantony:
this should solve #2088 
I haven't tested set and rise. Do we have tests for that? I am glad to see set and rise back. I have name supports sunset, sunrise, Pacifica,  and Tokyo 

https://libreswan.org/wiki/images/f/f1/Testnet-202102.png


cagney:
I suspect the thing to do is split systemd/* into separate bits for hostnamer and networks.  Then the base domains can just pull in the network stuff while the other linux domains (minus apline) can pull in both.

Issue #2148, opened 2025-04-01
stf_status vs _Bool


ueno:
As it seems like a common pattern that `true` or `false` is returned where `stf_status` is expected (and vice versa), I wrote a small script using libclang attached below.

If I run the following commands on Fedora 41, it detects:
```console
$ sudo dnf install python3-clang bear clang
$ bear -- make -j$(nproc)
$ python3 .../check-retval.py --clang-resource-dir="$(clang -print-resource-dir)"
ikev2_ike_sa_init.c:465:record_v2_IKE_SA_INIT_request
ikev2_ike_sa_init.c:532:record_v2_IKE_SA_INIT_request
ikev2_ike_intermediate.c:359:initiate_v2_IKE_INTERMEDIATE_request
ikev2_ike_intermediate.c:544:process_v2_IKE_INTERMEDIATE_request
ikev1_xauth.c:899:modecfg_send_request
ikev1_aggr.c:776:aggr_inR1_outI2
```

The ones in ikev2_ike_intermediate.c are being addressed in #2145, while the others look still valid. Maybe good to have this (or similar) in the CI to avoid that in the future?

[check-retval.py.gz](https://github.com/user-attachments/files/19545625/check-retval.py.gz)

cagney:
(Here CI == continuous integration)

> As it seems like a common pattern

as in common mistake :-(  I wish the compiler proper checked this.

I'm not sure where it should go.  There's:
- github's CI
Personally, I'm likely to ignore that
- testsuite's check-NN-*
up until now it hasn't looked at sources
- a bonus part of the test build
but test namespaces do things different so might skip it
- an addition to the default build
technically the most correct, but would require clang?
- the, I suspect defunct, `make check` target
I suspect it's defunct as I don't know that it is run

cagney:
> I don't think it's a good idea to integrate this into the standard build process though, as the check is a bit costly. The invocation like [scan-build](https://clang.llvm.org/docs/analyzer/user-docs/CommandLineUsage.html#scan-build) would be more suitable for this kind of check.

True.  I do think it needs to be hooked into the release engineering process somehow though.

PR #2149, opened 2025-04-02
testing: add basic GitHub workflow for static analysis

ueno:
This adds a script to automatically check stf_status/_Bool mismatch and plumbs it into a GitHub CI workflow.

Fixes: #2148

### DIFF
--- a/.github/check-retval.py
+++ b/.github/check-retval.py
@@ -1,0 +1,68 @@
+#!/usr/bin/python
+
+import argparse
+import clang.cindex
+import os.path
+import sys
+
+def walk(tu, node, results, func):
+    if node.kind == clang.cindex.CursorKind.FUNCTION_DECL:
+        func = node.spelling
+    elif node.kind == clang.cindex.CursorKind.RETURN_STMT:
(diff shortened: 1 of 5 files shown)


cagney:
Can the script live in `.github/workflows/` for now.  Once we've figured out how to integrate this into the build/release process we'll hopefully also figure out a better location.

Based on IRC discussion I'm going to delete `scripts/`.

The-Mule:
I think it would be useful in general to perhaps keep it in Github Actions even if it is integrated into build/release process. It might be handy for PRs of outside contributors.

cagney:
> I think it would be useful in general to perhaps keep it in Github Actions 

Yes.  Just waiting for a file shuffle, as scripts/ was deleted.
(having a github action pull in a test result would also be useful)

cagney (changes requested):
for moment script can live in .github/workflows/, scripts/ directory was deleted

ueno:
> for moment script can live in .github/workflows/, scripts/ directory was deleted

Done

PR #2150, opened 2025-04-02
testing: fix CodeQL build failure

ueno:
libcurl3-nss-dev is no longer provided in Ubuntu 24.04: https://packages.ubuntu.com/source/noble/curl


cagney:
well spotted (and timely as well since pluto no longer links against curl, instead _import_crl does)

PR #2167, opened 2025-04-18
packaging: add systemd-tmpfiles for directories under /var

jmarrero:
This is in service of image-mode systems in the Fedora/RHEL ecosystem.

I am not sure about the change of initsystems/systemd/libreswan.conf.in to add these directories or if I just should introduce a tmpfile to be individually packaged inside the fedora and rhel directories.

### DIFF
--- a/initsystems/systemd/libreswan.conf.in
+++ b/initsystems/systemd/libreswan.conf.in
@@ -1,1 +1,3 @@
 d @@RUNDIR@@ 755 root root -
+d /var/lib/ipsec 0700 root root -
+d /var/lib/ipsec/nss 0700 root root -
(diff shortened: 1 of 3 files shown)


paulwouters:
I think this change is fine.
I am a bit worried that pluto might start without a valid nss db in /var/lib/ipsec/nss
Specifically, the ipsec.service file contains "ipsec checknss" which needs the directories to be present:
```
ipsec checknss --nssdir /tmp/x/y
ERROR: destination directory "/tmp/x/y" is missing or permission denied
```

bleve:
There are issues. There is no need to change %files section for adding directories to tmpfiles. So package install can just fine create directories, tmpfiles are only needed to make sure directories do exist after boot.

Actually, with this change, nss directory must move out from /var to /usr, just like rpmdb moved from /var/lib/rpm to /usr/lib/rpm .

bleve:
Thank you for your patch, but there were issues and so it was rejected.

There is no need to have %ghost for files/directories which are created by package install.

Also disabling verify for owner and group is wrong: %verify(not owner group) - directories must be owned by root and must be verified for that.

There was no need to have restricted permissions for /var/lib/ipsec directory so I simplified tmpfiles part to one line and used correct makefile transofrmation macro  instead of hardcoded value.

So this was implemented slightly differently with commit bee7f36d480dd3912a5135a933b72025c819bcbf - hope this implemented what was needed.

PR #2322, opened 2025-06-25
whack: support structured logging in JSON

ueno:
This adds a `--json` option to "ipsec trafficstatus" to control whether the output is formatted in JSON.  Internally, this extends the jambuf interface to be called in a structural way, so the same code paths can be used for both plaintext and JSON formatted output, for example:
```c
  jam_array_start(buf);
    jam_object_start(buf);
      jam_field_start(buf, "field");
        jam_value_string(buf, "%s", "value");
      jam_field_end(buf");
    jam_object_end(buf);
  jam_array_end(buf);
```

### DIFF
--- a/include/whack.h
+++ b/include/whack.h
@@ -489,6 +489,8 @@ struct whack_message {
 	/* what metric to put on ipsec routes */
 	int metric;
 
+	bool whack_json;
+
 	/* space for strings (hope there is enough room) */
 	size_t str_size;
 	unsigned char string[4096];
--- a/programs/ipsec/ipsec.in
+++ b/programs/ipsec/ipsec.in
@@ -493,6 +493,10 @@ ipsec_whack() {
 		utc="${1}"
 		shift
 		;;
+	    --json)
+		json="${1}"
+		shift
+		;;
 	    --verbose)
 		verbose="${1}"
 		shift
@@ -509,7 +513,7 @@ ipsec_whack() {
 		;;
 	esac
     done
-    ${dry_run} "${IPSEC_EXECDIR}/whack" --ctlsocket "${CTLSOCKET}" ${asynchronous} ${conn_name} ${utc} ${cmd} ${remote_host}
+    ${dry_run} "${IPSEC_EXECDIR}/whack" --ctlsocket "${CTLSOCKET}" ${asynchronous} ${conn_name} ${utc} ${cmd} ${remote_host} ${json}
     exit $?
 }
 
--- a/programs/pluto/Makefile
+++ b/programs/pluto/Makefile
@@ -87,6 +87,8 @@ OBJS += ikev2_eap.o
 
 OBJS += state_db.o
 OBJS += show.o
+OBJS += show_text.o
+OBJS += show_json.o
 OBJS += binlog.o
 OBJS += retransmit.o
 OBJS += quirks.o
--- a/programs/pluto/ikev2_liveness.c
+++ b/programs/pluto/ikev2_liveness.c
@@ -140,7 +140,7 @@ static bool recent_last_contact(struct child_sa *child,
  * that half the channel is working.  All the incoming packets could
  * be retransmits.
  *
- * note: this mutates *st by calling get_sa_bundle_info
+ * note: this mutates *st by calling get_ipsec_traffic
  */
 
 void event_v2_liveness(struct state *st)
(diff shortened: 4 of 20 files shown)


ueno:
Here are sample outputs:

With `ipsec trafficstatus`:
```console
#2: connection: "mytunnel", type: ESP, add_time: 1750892751, inBytes: 700, outBytes: 240, maxBytes: 8Ei, id: @east
```

With `ipsec trafficstatus --json`:
```console
{"#2": {"connection": "\"mytunnel\"", "type": "ESP", "add_time": 1750892751, "inBytes": 700, "outBytes": 240, "maxBytes": "8Ei", "id": "@east"}}
```

cagney:
This patch is implementing the builder pattern.

If your interested in seeing how far this can be pushed, have a look at GDB and its ui-out code.  The display code can build either CLI or JSON-like output (this pre-dates JSON).  It was originally written in C.

A standard implementation has a builder super-class that has sub-classes for each output syntax - here CLI and JSON.
For libreswan, I suspect `struct show` would be a better starting point for the super class.  The `struct show` object is already fed lines of output, which means that it could build:

```
[
{ "state": 2:, "connection": "\"mytunnel\"", "type": "ESP", "add_time": 1750892751, "inBytes": 700, "outBytes": 240, "maxBytes": "8Ei", "id": "@east" },
{ "state": 3, "connection": "\"my-thertunnel\"", "type": "ESP", "add_time": 1750892751, "inBytes": 700, "outBytes": 240, "maxBytes": 112312389712038910892371089237089, "id": "@east" }
]
```

ueno:
Thank you; yes, `struct show` looks like the best place to implement this and separating concrete implementations by subclassing makes sense. I'll try to rework this PR along these lines (after rebasing #1830).

cagney:
@letoams what are your thoughts about this as the approach?  We've two alternatives:

- this where the status commands can build either of the output
- having separate commands for showing text and json

both have trade offs.  For instance, where not changing the output is critical, preserving the text code may be bettter VS a comand like `ipsec connectionstatus` where having to only have an option in one place is better.  I suspect we'll end up with both.

ueno:
@cagney Sorry for the delay; I've updated the PR to incorporate your suggestion:

- moved the frontend interface from `struct jambuf` to `struct show`
- separated the backend implementations into `show_text.c` and `show_json.c`, which implement vtable defined in `show_ops.h`

The result is mostly the same as before. With `ipsec trafficstatus`:
```console
serial=#2, connection="mytunnel", type=ESP, add_time=1758869941, inBytes=8180, outBytes=7724, maxBytes=2^63B, id=@east
serial=#3, connection="mytunnel", type=ESP, add_time=0, inBytes=35508, outBytes=33740, maxBytes=2^63B, id=@east
```
while `ipsec trafficstatus --json` produces:
```console
{"serial": "#2", "connection": "\"mytunnel\"", "type": "ESP", "add_time": 1758869941, "inBytes": 8180, "outBytes": 7724, "maxBytes": "2^63B", "id": "@east"}
{"serial": "#3", "connection": "\"mytunnel\"", "type": "ESP", "add_time": 1758869949, "inBytes": 35508, "outBytes": 33740, "maxBytes": "2^63B", "id": "@east"}
```
The format is conceptually based on the "structured logging", i.e., key-value pairs.

ueno:
A couple of caveats:

1. The output of `ipsec trafficstatus` is slightly different from the original (as the log lines consist of key-value pairs)
2. `struct show` provides two different interfaces: the original line-based one vs the structured logging, which are mutually exclusive
3. To complete the feature, `ipsec status` should probably be updated to using the structured logging interface as well

ueno:
After a second thought, I realize it might be better to define the vtable (show_ops) at higher level. Currently it is:
```c
struct show_ops {
	void (*raw_va_list)(struct jambuf *buf, const char *message, va_list ap);
	void (*string_va_list)(struct jambuf *buf, const char *message, va_list ap);
	void (*member_start)(struct jambuf *buf, const char *name);
	void (*member_end)(struct jambuf *buf);
	...
};
```
which could be:
```c
struct show_ops {
	void (*show_kernel_interface)(struct show *s);
	void (*show_traffic_status)(struct show *s);
	...
}
```
That way, we can keep the compatibility while giving the text output more flexibility. A downside is that there will be a significant amount of duplicated code, and if you add a new field to the output, you would need to edit two places. @cagney I suspect this was what you meant before?

Issue #2356, opened 2025-07-21
cannot create /etc/resolv.conf Permission denied


thearcher13:
### Please Tell Us About Your Operating System

22.04.1-Ubuntu

### Please Describe Your IPsec Configuration

config setup     
   ikev1-policy=accept
        protostack=netkey
        uniqueids=no
        dnssec-enable=no
conn part
        authby=secret
        auto=add
        rekey=no
        pfs=yes
        aggressive=yes
        keyexchange=ikev1
        cisco-unity=no
        cisco-split=yes
        leftsubnet=0.0.0.0/0
        rightsubnet=0.0.0.0/0
        ike = aes256-sha256;modp2048
        esp = aes256-sha256;modp2048
        left=%defaultroute
        leftid=@[id]
        leftxauthclient=yes
        leftmodecfgclient=yes
        leftxauthusername="[User}"
        modecfgpull=yes
        right=[domain address]
        rightxauthserver=yes
        rightmodecfgserver=yes

### Please Tell Us What Happened?

i can connect to the tunnel but tunnel's dns doesn't get applied and when i connect to the tunnel i see permission denied even that i start the tunnel with sudo. i tried starting the tunnel with root user as well and i still had this problem

### Relevant log output

```shell
#2: up-client output: updating resolvconf
#2: up-client output: /usr/local/libexec/ipsec/_updown.xfrm: 427: cannot create /etc/resolv.conf: Permission denied
```

letoams:
On Mon, 21 Jul 2025, Amirreza wrote:

> #2: up-client output: updating resolvconf
> #2: up-client output: /usr/local/libexec/ipsec/_updown.xfrm: 427: cannot create /etc/resolv
> .conf: Permission denied

Possibly someone / something made /etc/resolv.conf immutable?

What is managing /etc/resolv.conf on your system? Perhaps this is
networkmanager and we should use nmcli to update the dns. Perhaps
it is systemd-resolved and then something (I dont know what) needs
to happen to reconfigure systemd-resolved ?

Paul

thearcher13:
thanks for the reply

i checked the file:
This is /run/systemd/resolve/resolv.conf managed by man:systemd-resolved(8).

thearcher13:
/etc/resolv.conf is a symlink to /run/systemd/resolve/resolv.conf and its not immutable

lsattr /run/systemd/resolve/resolv.conf                                                                                                                            
---------------------- /run/systemd/resolve/resolv.conf

[...]

PR #2379, opened 2025-07-31
added systemd-resolved DNS support

thearcher13:
it was asked in:
https://github.com/libreswan/libreswan/issues/2356
also fixed the problems that were mentioned in 
https://github.com/libreswan/libreswan/pull/2376

### DIFF
--- a/programs/_updown.xfrm/_updown.xfrm.in
+++ b/programs/_updown.xfrm/_updown.xfrm.in
@@ -345,26 +345,26 @@ updateresolvconf() {
     if command -v resolvectl >/dev/null 2>&1 && systemctl is-active --quiet systemd-resolved && grep -q '127\.0\.0\.53' "${ETC_RESOLV_CONF}"; then
         echo "[updown.xfrm] Detected systemd-resolved, applying DNS via resolvectl"
 
-	if [ ! -f "${DNS_BACKUP}" ]; then
-		resolvectl dns "${PLUTO_INTERFACE}" | sed -E 's/^Link[^:]*: *//' | grep -v '^[[:space:]]*$' | xargs > "${DNS_BACKUP}" 2>/dev/null
+	if [ ! -f "$DNS_BACKUP" ]; then
+		resolvectl dns "$PLUTO_INTERFACE" | sed -E 's/^Link[^:]*: *//' | grep -v '^[[:space:]]*$' | xargs > "$DNS_BACKUP" 2>/dev/null
 	else
 		echo "[updown.xfrm] DNS backup already exists, skipping backup"
 	fi
 
-	if [ ! -f "${DOMAIN_BACKUP}" ]; then
-		resolvectl domain "${PLUTO_INTERFACE}" | sed -E 's/^Link[^:]*: *//' | grep -v '^[[:space:]]*$' | xargs > "${DOMAIN_BACKUP}" 2>/dev/null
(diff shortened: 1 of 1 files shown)


cagney:
we'll need to work out testing
- I'll finish adding alpine to the testsuite; that gives us a VM with no systemd; probably in a fortnight
- I guess we'll need to hack a test so it enables systemd's dns before starting pluto

letoams:
I think the systemctl call is too expensive. And there is no guarantee glibc isn’t trying to use it via nssswitch.conf either or is not using it while the service runs. Systemd-resolved is a mess

thearcher13:
i've tested again on ubuntu with change in the code and when systemd resolved was managing the dns it worked fine. i also tried when systemd-resolved disabled and fallback worked fine. 

i think If there is no systemd at all
`command -v resolvectl` will fail

for the glibc that was mentioned, i dont know much about it. 

what do you suggest? do you think is there any better condition to handle this and make the "if" statement better or is this one alright 

@cagney @letoams

thearcher13:
there was a problem when restoring the dns for systemd resolved which is fixed now.

thearcher13:
also improved logic of determining if systemd resolved is handling dns by adding the check that "127.0.0.53" exists in /etc/reolv.conf

thearcher13:
@cagney @letoams

cagney:
(we brainstormed this F2F), the suggestion is:

- add an option to `ipsec.conf` `config setup` to specify the DNS being used
for instance, `dns=systemd|networkmanager|file|unbound|auto`, 
initially it could be `systemd` and `file` (default)
 `auto`, which would probe the system for the answer, could be added later
- pass the setting, as an environment variable, to `updown`
this way the need to re-probe is removed; and the behaviour is predictable and configurable

(there's a suggestion to run DNS updates separately but that is way beyond the scope of this change)

@letoams?

cagney:
as an aside, this:
```
static err_t read_sysctl(const char *ctl, uintmax_t *value, struct verbose verbose)
{
        const char *sysctl[] = {
                "sysctl",
                "-n",
                ctl,
                NULL,
        };
        struct server_run result = server_runve_chunk(sysctl, NULL, null_shunk, verbose);
        if (result.status != 0) {
                return "sysctl exited with a non-zero status";
        }
        if (result.output.len == 0) {
                return "sysctl exited with no output";
        }
        shunk_t shunk = HUNK_AS_SHUNK(result.output);
        err_t e = shunk_to_uintmax(shunk, &shunk, 10, value);
        if (e != NULL) {
                return "sysctl output is non-numeric";
        }
        free_chunk_content(&result.output);
        return NULL;
}
```
illustrates how to directly invoke (i.e., not via a shell) and then capture the output from a program

cagney:
@cagney wonders if there's anything internal to pluto that needs to know the System's DNS resolver.

cagney:
the config-setup variable is `dns-resolver={file,systemd}` and the `updown` variable is `PLUTO_DNS_RESOLVER=`

cagney:
@thearcher13 would you like me to update/merge this so it uses `PLUTO_DNS_RESOLVER=`?

thearcher13:
Hi, sorry for the late reply. sure. @cagney

cagney:
I've merged this

cagney:
@thearcher13 fyi, the only tweaks needed from your work were:

- test case
- move code around and add switch

nice work

PR #2415, opened 2025-08-27
testing: update KVM_LINUX_ISO_CKSUM to the latest

ueno:
Without this, running "kvm install" fails with:
```
  + echo 'SHA256 (/home/ueno/devel/pool/Fedora-Server-dvd-x86_64-41-1.4.iso.tmp) = 32d9ab1798fc8106a0b06e873bdcd83a3efea8412c9401dfe4097347ed0cfc65'
  + cksum -c /home/libreswan/devel/pool/Fedora-Server-dvd-x86_64-41-1.4.iso.tmp: FAILED cksum: WARNING: 1 computed checksum did NOT match
```
The checksum is actually
6037e489103401a6ad4e54a4bcb2df7525693bdc3f2ce4aa895838b65647e551 according to:
https://download.fedoraproject.org/pub/fedora/linux/releases/41/Server/x86_64/iso/Fedora-Server-41-1.4-x86_64-CHECKSUM


cagney:
Ah, it was broken a combination of dfbba59cb020dfe4402dcb63c7373bb11fd5b87f and me already having the ISO downloaded.  tks

PR #2416, opened 2025-08-27
crypto: use the ML-KEM mechanisms standardized in PKCS#11 3.2

ueno:
The latest (unreleased) version of NSS has support for the ML-KEM mechanisms from the final PKCS#11 3.2 standard:
https://phabricator.services.mozilla.com/D259698

Use them instead of the vendor-specific mechanisms for future proof, as some of them (e.g., CKM_KYBER) will be no-op in RHEL.

Marking draft as this needs to wait for a next NSS release (3.116).

### DIFF
--- a/lib/libswan/ike_alg_kem_ml_kem_nss_ops.c
+++ b/lib/libswan/ike_alg_kem_ml_kem_nss_ops.c
@@ -26,13 +26,31 @@
 #include "passert.h"
 #include "lswalloc.h"
 
+/*
+ * The official ML-KEM mechanisms from PKCS#11 3.2 are only supported
+ * in NSS 3.116 or later. Use the vendor-specific constants otherwise.
+ */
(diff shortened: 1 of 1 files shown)


ueno:
cc @rjrelyea

cagney:
@cagney gently weeps

I guess the top of the file will need some sort of garish
```
#if NSS_VERSION < magic_number
#define NEW_NAME OLD_NAME
#endif
```
at the top.  The alternative, of two files with subtly different implementations is clearly overkill.

letoams:
Since the window of unofficial nss support is so small, why not skip it and only support it for nss 3.116+  ?

cagney:
debian

ueno:
@cagney I added #ifdefs.

cagney:
an aside, the debian being used by github's build farm is really old

PR #2418, opened 2025-08-28
logging: fix str_nss_oid

ueno:
If the OID is known, str_nss_oid previously returned an uninitialized buffer, instead of the actual description.

I stumbled upon this when testing with `update-crypto-policies --set DEFAULT:TEST-PQ`, where SHA-1 is disabled by policy; the log contained `SGN_Digest(...gibberish...) failed`.


cagney:
big oops; pushed, thanks

PR #2439, opened 2025-09-17
testing: make check-retval.py work with relative pathnames

ueno:
The check-retval GitHub action had been broken for a while, after the build infrastructure started using relative paths of C source files instead of the absolute paths.


cagney:
cherry-picked; thanks!

PR #2501, opened 2025-10-14
building linux: turn on USE_XFRM_HEADER_COPY for Ubuntu

vukasink:
Additionally, set USE_ML_KEM_768 to false. nss3 on Ubuntu seems not to have new NSS version (>= 3.105 ?) that support USE_ML_KEM_768 (manually tested on Ubuntu 22.04 and Ubuntu 24.04, the build doesn't work)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,6 @@ For Debian / Ubuntu / Mint
       libpam0g-dev libcap-ng-dev libldns-dev xmlto \
       libcurl4-openssl-dev
 
-apt-get install build-essential pkg-config \
-      bison fles
-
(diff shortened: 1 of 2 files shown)


vukasink:
Piggybacking the second commit on this PR. It seems there was unintentional typo in README.md

vukasink:
Piggybacking the third commit on this PR. It seems like libnss3-tools is required for Debian(-like). I had to install it today on both Raspberry Pi OS (Debian 13-based) and Ubuntu 24.04. Otherwise, certutil does not exist and ipsec start fails.

cagney:
nice!  will drop right in

Issue #77, opened 2016-11-23
Add libgmp3-dev to dependencies list


roadriverrail:
Looks like libgmp3 is now a dependency; it should be added to the package install commands in your README.md.

bleve:
No. libgmp is not build dependency.

letoams:
libgmp is no longer used by libreswan

roadriverrail:
Hunh.  I was getting a build break until I installed it.  My mistake,
clearly.

letoams:
Then you are using an older version of libreswan

letoams:
gmp is no longer used by libreswan

PR #2595, opened 2026-01-23
Clear parent references when deleting Child SA

The-Mule:
If CREATE_CHILD_SA response happens after larval child is gone process_v2_CREATE_CHILD_SA_child_response() accesses already freed child (through parent's wip_sa), found by @igsilya:

```
Thread 1 "pluto" received signal SIGSEGV, Segmentation fault.
0x000055964aec8718 in set_larval_v2_transition (larval=0x55966e031650,
    to=0x55964afea1a0 <state_v2_ESTABLISHED_CHILD_SA>,
    where=where@entry=0x55964afd62a0 <here>) at ikev2_child.c:86
86              const struct v2_transition *transition =
(gdb) #0  0x000055964aec8718 in set_larval_v2_transition (larval=0x55966e031650,
    to=0x55964afea1a0 <state_v2_ESTABLISHED_CHILD_SA>,
    where=where@entry=0x55964afd62a0 <here>) at ikev2_child.c:86
#1  0x000055964ae1244f in process_v2_CREATE_CHILD_SA_child_response (
    ike=0x55966dff5000, larval_child=<optimized out>,
    response_md=0x55966e0463b0) at ikev2_create_child_sa.c:1426
#2  0x000055964aec55ff in v2_dispatch (ike=ike@entry=0x55966dff5000,
    md=md@entry=0x55966e0463b0,
    svm=svm@entry=0x55964affb758 <v2_CREATE_CHILD_SA_response_transition+120>)
    at ikev2.c:1340
```

Crash happens in:

https://github.com/libreswan/libreswan/blob/3d59af8308a5cb2f6d777f6ec9c0ef8b095f7b89/programs/pluto/ikev2_child.c#L82-L93

when accessing `larval->sa.st_state->v2.larval_sa_transition` and we realized that it might happen that larval child gets deleted during CREATE_CHILD_SA exchange. It then points to `&state_undefined` (without `v2`) and crash happens. The proper fix seems to be to clear parent IKE SA reference to the larval child when it is being deleted so that `process_v2_CREATE_CHILD_SA_child_response()` won't get to `set_larval_v2_transition()`.

https://github.com/libreswan/libreswan/blob/3d59af8308a5cb2f6d777f6ec9c0ef8b095f7b89/programs/pluto/ikev2_create_child_sa.c#L1431-L1450

_Update (March 3): Proper solution seems to be to defer child deletion._

### DIFF
--- a/programs/pluto/ikev2_create_child_sa.c
+++ b/programs/pluto/ikev2_create_child_sa.c
@@ -1445,6 +1445,20 @@ stf_status process_v2_CREATE_CHILD_SA_child_response(struct ike_sa *ike,
 	pexpect(larval_child->sa.st_sa_kind_when_established == CHILD_SA);
 
 	/*
+	 * If the connection of larval child gets deleted while waiting for the
+	 * CREATE_CHILD_SA response, childs deletetion is deferred and now is the
+	 * time to delete it properly.
+	 */
+	struct connection *c = larval_child->sa.st_connection;
+	if (!c->policy.up && !c->policy.keep) {
+		llog_sa(RC_LOG, larval_child,
+			"connection was deleted while waiting for CREATE_CHILD_SA response, sending DELETE");
+		submit_v2_delete_exchange(ike, larval_child);
+		ike->sa.st_v2_msgid_windows.initiator.wip_sa = NULL;
+		return STF_OK;
+	}
+
+	/*
 	 * Drive the larval Child SA's state machine.
 	 */
 	set_larval_v2_transition(larval_child, &state_v2_ESTABLISHED_CHILD_SA, HERE);
@@ -2173,7 +2187,21 @@ stf_status process_v2_CREATE_CHILD_SA_failure_response(struct ike_sa *ike,
 		return STF_INTERNAL_ERROR;
 	}
 
-        pstat_sa_failed(&(*larval_child)->sa, REASON_TRAFFIC_SELECTORS_FAILED);
+	/*
+	 * If the connection of larval child gets deleted while waiting for the
+	 * CREATE_CHILD_SA response, childs deletetion is deferred and now is the
+	 * time to delete it properly.
+	 */
+	struct connection *c = (*larval_child)->sa.st_connection;
+	if (!c->policy.up && !c->policy.keep) {
+	  llog_sa(RC_LOG, (*larval_child),
+		  "connection was deleted while waiting for CREATE_CHILD_SA response, sending DELETE");
+	  submit_v2_delete_exchange(ike, *larval_child);
+	  ike->sa.st_v2_msgid_windows.initiator.wip_sa = NULL;
+	  return STF_OK;
+	}
+
+	pstat_sa_failed(&(*larval_child)->sa, REASON_TRAFFIC_SELECTORS_FAILED);
 
 	stf_status status = STF_ROOF; /*IKE;place holder*/
 
--- a/programs/pluto/routing.c
+++ b/programs/pluto/routing.c
@@ -1123,6 +1123,31 @@ static bool teardown_child_dispatch_ok(struct connection *c,
 
 static void teardown_child_post_op(const struct routing_annex *e)
 {
+	struct child_sa *child = *e->child;
+	struct state *st = &child->sa;
+
+	/*
+	 * If (cuckoo) larval child has CREATE_CHILD_SA request outstanding, its
+	 * deletion is deferred until the response arrives or until its timeout.
+	 */
+	if (st->st_clonedfrom != SOS_NOBODY && !IS_CHILD_SA_ESTABLISHED(st)) {
+		struct ike_sa *ike = ike_sa_by_serialno(st->st_clonedfrom);
+		if (ike != NULL && ike->sa.st_v2_msgid_windows.initiator.wip_sa == child && ike->sa.st_viable_parent) {
+			llog(RC_LOG, st->logger,
+				"deletion of larval child #%u for connection '%s' deferred (pending CREATE_CHILD_SA response)",
+				st->st_serialno, st->st_connection->name);
+
+			/*
+			 * Since we deferred deletion we also need to restore negotiating_child_sa
+			 * that was cleared by teardown. With that weeding visit won't attempt to
+			 * delete it again.
+			 */
+			st->st_connection->negotiating_child_sa = st->st_serialno;
+			*e->child = NULL;
+			return;
+		}
+	}
+
 	delete_child_sa(e->child);
 }
 
--- a/programs/pluto/terminate.c
+++ b/programs/pluto/terminate.c
@@ -484,6 +484,17 @@ void terminate_and_delete_connections(struct connection *c,
 		 */
 		whack_attach(c, logger);
 		terminate_and_down_and_unroute_connections(c, where);
+
+              /* connection has child sa with deferred deletion */
+              if (c->negotiating_child_sa != SOS_NOBODY) {
+                     ldbg(c->logger, "connection '%s' has children deferring deletion",
+                         c->name);
+                     /* we detach whack even though deletetion is deferred */
+		       whack_detach(c, logger);
+		       terminate_and_down_and_unroute_connections(c, where);
+                     return;
+              }
+
 		/* leave whack attached during death */
 		connection_delref(&c, logger);
 		return;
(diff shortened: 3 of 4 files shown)


cagney:
why did the larval child disappear?

cagney:
10 points to @igsilya

The-Mule:
It can happen that CREATE_CHILD_SA response arrives after child gets deleted.

There's actually one more thing to address:

https://github.com/libreswan/libreswan/blob/3d59af8308a5cb2f6d777f6ec9c0ef8b095f7b89/programs/pluto/ikev2_create_child_sa.c#L1440-L1443

`STF_INTERNAL_ERROR` seems to be too strict, doesn't it? Since it can happen that larval child sa is deleted at this point, it might be okay to handle it grafeully (ie. STF_OK?).

Here's what we saw:

1. Child SA `#77` created and CREATE_CHILD_SA request sent (Child SA `#77` becomes `sa.st_v2_msgid_windows.initiator.wip_sa` of IKE SA `#70`)
```
20:02:14.668485: "tun-20-out-1" #77: initiating Child SA using IKE SA #70
20:02:14.673856: | "tun-20-out-1" #77: adding larval SA to IKE SA #70 message initiator queue; sec_label=
20:02:14.673952: "tun-20-out-1" #77: adding CREATE_CHILD_SA request to IKE SA #70's message queue
20:02:14.675457: "tun-20-out-1" #77: sent CREATE_CHILD_SA request to create Child SA using IKE SA #70 {ESP <0x5c5eff1c}
```

2. Child SA `#77` cuckoo to "tun-20-in-1" and its original "tun-20-out-1" terminated
```20:02:14.717105: | visit:   dispatch negotiating principal Child SA #77 (cuckoo) with another connection's established IKE SA "tun-20-in-1" #70 (cuckold)
20:02:14.717108: "tun-20-out-1": terminating SAs using this connection
14 20:02:14.775595: | "tun-20-out-1" #77: .st_on_delete.skip_send_delete no->true (delete_child_sa() +747 programs/pluto/state.c)
20:02:14.775641: | "tun-20-out-1" #77: delete_state() skipping log_message:no
20:02:14.775647: | "tun-20-out-1" #77: .st_on_delete.skip_log_message no->true (delete_state() +878 programs/pluto/state.c)
20:02:14.775652: | pstats #77 ikev2.child deleted other
```
Child SA `#77` now gone (`&state_undefined`) but IKE SA `#70` `sa.st_v2_msgid_windows.initiator.wip_sa` still points to it.

3. CREATE_CHILD_SA response received
```
20:02:14.806933: |    Notify Message Type: v2N_USE_TRANSPORT_MODE (0x4007)
20:02:14.806958: | selected state microcode process Child SA response (new or rekey) (CREATE_CHILD_SA)
20:02:14.806964: | #70.st_v2_transition ESTABLISHED_IKE_SA (...) -> ESTABLISHED_IKE_SA (... process Child SA response (new or rekey) (CREATE_CHILD_SA))
20:02:14.806985: | "tun-20-in-1" #70: Message ID: initiator starting response 0; initiator .wip=-1->0
20:02:14.806988: | calling processor process Child SA response (new or rekey) (CREATE_CHILD_SA)
20:02:14.806994: EXPECTATION FAILED: larval_child->sa.st_sa_kind_when_established == CHILD_SA (process_v2_CREATE_CHILD_SA_child_response() +1421 programs/pluto/ikev2_create_child_sa.c)
```

4. Crash `process_v2_CREATE_CHILD_SA_child_response()` -> `set_larval_v2_transition()` -> `larval_child = ike->sa.st_v2_msgid_windows.initiator.wip_sa` (NULL) -> `pexpect(larval_child->sa.st_sa_kind_when_established == CHILD_SA)` -> SIGSEGV.

cagney:
[Simultaneous Child SA Rekeying](https://www.rfc-editor.org/rfc/rfc7296.html#section-2.8.1) makes two points:

- both SPIs are used when identifying the redundant child
the initiator only has this information once it receives the CREATE_CHILD_SA response
- the redundant child is deleted using a new exchange, i.e., **after** the CREATE_CHILD_SA completes
this keeps the two ends consistent

to implement this the redundant child sa needs to linger until the delete exchange completes

I'm not sure that this crossing stream case should be treated differently (the only difference being the is-this-redundant test).

The-Mule:
Right, if I read that correctly then what we want is not to delete child SA when it becomes cuckoo but instead keep it hanging there until CREATE_CHILD_SA response arrives and delete it when response is processed, ie. in `process_v2_CREATE_CHILD_SA_child_response()`. This approach won't clash with 

https://github.com/libreswan/libreswan/blob/3d59af8308a5cb2f6d777f6ec9c0ef8b095f7b89/programs/pluto/ikev2_create_child_sa.c#L1440-L1443

because `larval_child` will only get deleted after this check...

cagney:
yes, it should also stop the original crash

cagney:
Here's a table to play with for the initiator end:

|event|best nonce|connection state|connection owner|kernel state/policy|comments|
|-|-|-|-|-|-|
|send CREATE_CHILD_SA request|-|*_NEGOTIATION|local|possible negotiation policy|-|
|receive CREATE_CHILD_SA request|-|*_NEGOTIATION|local|no change|Least worst option?<br>Can't yet compare NONCEs<br>`so_serial_t .st_my_nemisis`
|receive CREATE_CHILD_SA response|local|ROUTED_TUNNEL|local|install local<br>state/policy|peer deletes its new SA<br>(local deletes rekeyd SA)
|receive CREATE_CHILD_SA response|peer|>ROUTED_TUNNEL|>peer|install peer<br>state/policy|local deletes its new SA<br>(peer deletes rekeyed SA)

Least worst option?  There seems to be two choices:

- do nothing until CREATE_CHILD_SA response arrives
- switch connection to peer's state/policy

The RFC says to _just note the situation and respond as usual_
- suggesting do nothing until CREATE_CHILD_SA response arrives

But then, right at the end, it has _When A receives this [CHILD_SA_NOT_FOUND] error, it already knows there was simultaneous rekeying, so it can ignore the error message._
- mostly, it needs to install the peer's state policy

cagney:
Here's an interesting variation:

- local initiates, sends CREATE_CHILD_SA request to peer
- local receives CREATE_CHILD_SA request from peer
per rfc, it's put on hold
response sent
- local receives N(DELETE) for peer's child
peer receives the local's CREATE_CHILD_SA request, and responded but response is lost
peer receives it's CREATE_CHILD_SA response, seen that it lost, and initiates delete
peer receives re-transmit of local's CREATE_CHILD_SA, re-transmits response
hence, messages are out of order
- local responds to delete of peer's child
- local receives CREATE_CHILD_SA response for its child
there's no longer a clash so local kernel state/policy can be installed

I think this gives weight to the argument that an end should delay installing the kernel state/policy until there's complete information

cagney:
This should be dropped as the PEXPECT isn't the problem?

The-Mule:
I want to take a closer look at this this week, sorry for the delay @cagney. I think the issue described here is still valid - there's a crash possible, but my original solution is wrong. It is definitely a very rare race condition but still I think it's something we can fix.

cagney:
> I think the issue described here is still valid

Yes, there's definitely a bug.  I was wondering more if you wanted to keep this pull request and update, or close it and start anew.

I'll mark it as "draft"/

The-Mule:
It looks like I might have a sort-of working version - general idea of this change is reviewable. In short this PR introduces so-called "deferred deletion" of larval child in specific situations:

- if it has CREATE_CHILD_SA request outstanding and
- if its parent IKE is viable.

Deletion is deferred to the point when:

- CREATE_CHILD_SA response arrives (then proper DELETE exchange happens) or
- the parent IKE is torn down or
- the timeout.

Also even if deletion is deferred we detach whack so that `whack --delete` won't get stuck. 

It's not the end though since some tests seems to be affected by the change and (e.g. ikev2-algo-17-no-sha1-01 but there are more of them) deferred larval child deletion is causing timeout waititing and make some tests slower. There's probably still more work needed.

cagney:
Are there two problems here?

- a create child sa being crossed by the peer sending a request to delete / replace / ... that same larval child sa
here, everything is still civil, the child can be marked non-viable (perhaps loosing ownership of the SA), and then, when the create-child-sa response comes back, be deleted

- a connection initiating a create_child_sa, simultaneous to that connection being forcefully deleted using whack
here, terminate is expected to wipe any of the connection's states before returning. If the connections a cuckoo, then the only option I see is to cut off the cuckold's head; if the cuckold isn't deleted it is going to be in an inconsistent state

cagney:
I'm going to change `ipsec delete ikev2-cuckoo` to delete the cuckold.

While my test is simpler than what we've got here - it just wedges the IKE SA during a CREATE_CHILD_SA exchange - I believe the change will fix this case, as per:
   pexpect a dead IKE SA to have no lingering larval children #2930
the current code leaves the .wip_sa non-NULL and pointing at a deleted state; and after the change it all will be gone

cagney:
I'm sure its fixed by b4c6fe8419d2f063fdf6e276d1bdcdf35c51f430

PR #2598, opened 2026-01-29
Fix compilation error on Fedora 44

ueno:
On Fedora 44, where -Wunused-but-set-variable is enabled with GCC 16, the build fails with:
```
  ikev2_proposals.c: In function ‘process_transforms’:
  ikev2_proposals.c:453:46: error: variable ‘local_proposal’ set but not used [-Werror=unused-but-set-variable=]
    453 |                 const struct ikev2_proposal *local_proposal;
        |                                              ^~~~~~~~~~~~~~
  ikev2_proposals.c:773:32: error: variable ‘local_proposal’ set but not used [-Werror=unused-but-set-variable=]
    773 |         struct ikev2_proposal *local_proposal;
        |                                ^~~~~~~~~~~~~~
```
This fixes it by adding the UNUSED attribute to the local variable.

### DIFF
--- a/programs/pluto/ikev2_proposals.c
+++ b/programs/pluto/ikev2_proposals.c
@@ -450,7 +450,7 @@ static int process_transforms(struct pbs_in *prop_pbs, struct jambuf *remote_jam
 	 */
 	{
 		int local_propnum;
-		const struct ikev2_proposal *local_proposal;
+		const struct ikev2_proposal *local_proposal UNUSED;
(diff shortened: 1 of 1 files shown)


cagney:
will appear shortly, tks!

ueno:
Thanks; now that it's [merged](https://github.com/libreswan/libreswan/commit/969765e2b922b009ad30d094f31edc81cb64de72), closing this.

Issue #2617, opened 2026-02-13
in fd.c, attach output to libevent


cagney:
### What needs to be done

so it can use non-blocking writes

Issue #1859, opened 2024-10-21
pluto lock up by ipsec status


igsilya:
### What happened?

If someone calls `ipsec status`, re-directing the `stdout` to a pipe and doesn't read from this pipe, or connects to `pluto.ctl` directly and doesn't read the reply from the socket, `pluto` will block on `sendmsg` and will not process any other requests.

This happens with `ovs-monitor-ipsec` daemon of Open vSwitch project. The daemon is parsing the output of `ipsec status` to figure out which connections are already loaded/active and loads ones that are not there. Due to format change in libreswan 5, the parsing code left some text unread in the pipe, so `ipsec status` got stuck fully blocking `pluto`. We can avoid that in `ovs-monitor-ipsec`, but it's still not a good thing that this can happen.  Ideally we would like to not parse the `stdout`, but there is no other interface today, and that's a separate topic.

**How to reproduce:**

1. Create a sufficiently large `ipsec.conf`.  In my case it contains about 80 connections.  This is needed to have a sufficiently large output for `ipsec status` command, so the pipe and the socket send/recv buffers are not large enough to hold the whole output.  Attached an archive [node-1.tar.gz](https://github.com/user-attachments/files/17464928/node-1.tar.gz) with directory with such a file and the initialized nss db.

2. Start `pluto`.
```
   # cd node-1
   # ipsec pluto --config $(pwd)/ipsec.conf --ipsecdir $(pwd)/      \
                 --nssdir $(pwd)/ipsec.d --logfile $(pwd)/pluto.log \
                 --secretsfile $(pwd)/secrets --rundir  $(pwd)/
```
3. Start the following python script:
```
   import subprocess
   import time

   proc = subprocess.Popen(['ipsec', 'status', '--ctlsocket', './pluto.ctl'],
                           stdout=subprocess.PIPE)
   time.sleep(3600)
```

5. Try to run `ipsec status` or any other `ipsec` command in a separate terminal.  Observe that `pluto` doesn't reply.

**Looking at the processes:**
```
# ps -aux | grep ipsec
root     1462372  Ssl  /usr/libexec/ipsec/pluto <...>
root     1463521  S+   /usr/bin/sh /usr/sbin/ipsec status --ctlsocket ./pluto.ctl
root     1463522  S+   /usr/libexec/ipsec/whack --ctlsocket ./pluto.ctl --status
```

**What each of these processes are doing:**

1. `ipsec status` just waits for the whack to finish:
```
   # strace -p 1463521
[...]

PR #2624, opened 2026-02-18
libswan: move libevent event_base to libswan

### LINKED ISSUE #2616: move libevent event loop in server.c to libswan
### What needs to be done

this is so that it is available to fd.c

Eshaan-byte:
Closes #2616

Adds a global_event_base get/set interface in lib/libswan/, following the existing global_logger pattern. Pluto's init_server() publishes the event_base after creation and free_server() clears it before freeing. get_pluto_event_base() now delegates to get_global_event_base().

The header uses only a forward declaration (struct event_base;) with no libevent headers, so non-pluto programs linking libswan.a are unaffected.

This makes the libevent event loop accessible from lib/libswan/ (e.g., fd.c) for future non-blocking/buffered I/O work (#2617, #1859).

New files:
-include/global_event_base.h — header declaring get_global_event_base() / set_global_event_base()
-lib/libswan/global_event_base.c — stores the global event_base pointer

Modified files:
-lib/libswan/Makefile — added global_event_base.o to OBJS
-programs/pluto/server.c — calls setter in init_server() / free_server(), delegates getter

### DIFF
--- a/include/global_event_base.h
+++ b/include/global_event_base.h
@@ -1,0 +1,25 @@
+/* global event_base declaration, for libreswan
+ *
+ * Copyright (C) 2025 Eshaan Gupta
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 2 of the License, or (at your
+ * option) any later version.  See <https://www.gnu.org/licenses/gpl2.txt>.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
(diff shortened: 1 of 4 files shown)


cagney:
Looking at the change.  I see the event base:
```
/* pluto's main Libevent event_base */
static struct event_base *pluto_eb =  NULL;
```
isn't moved (it's still in `server.c`).

If `pluto_eb` were moved, then some of the code in `server.c` would likely also need to be moved with it.  How much is something to consider.

Also, as I noted here https://github.com/libreswan/libreswan/issues/2616#issuecomment-3902856454 there might be better to leave event_base where it is, and instead pass a parameter.

btw, it's 2026.

cagney:
Closing; can be re-opened once updates are ready
(if this is GSoC related, then please see the guidelines)

Issue #2492, opened 2025-10-12
Error message "keyword auth, invalid value: rsa-sha2" should include all valid values in the error message


minfrin:
### Please Tell Us About Your Operating System

Linux localhost 5.14.0-570.42.2.el9_6.x86_64 https://github.com/libreswan/libreswan/issues/1 SMP PREEMPT_DYNAMIC Sun Sep 14 13:59:34 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
Libreswan 4.15
Peer is MacOS Sequoia.

### Please Describe Your IPsec Configuration

```
        leftauth=rsa-sha2
```

### Please Tell Us What Happened?

When following the man page as follows:

```
       leftauth
           How the security gateways will authenticate to the other side in the case of asymmetric authentication; acceptable values are rsasig or rsa for RSA Authentication with SHA-1, rsa-sha2 for
           RSA-PSS digital signatures based authentication with SHA2-256, rsa-sha2_384 for RSA-PSS digital signatures based authentication with SHA2-384, rsa-sha2_512 for RSA-PSS digital signatures based
           authentication with SHA2-512, ecdsa for ECDSA digital signatures based authentication, secret for shared secrets (PSK) authentication and null for null-authentication. There is no default
           value - if unset, the symmetrical authby= keyword is used to determine the authentication policy of the connection.
```

and using the value 'rsa-sha2', the server does not start and the error is as follows:

```
Oct 12 10:35:46 aurora.example.com addconn[451809]: ERROR: /etc/ipsec.d/ipv6-test.conf: 19: keyword auth, invalid value: rsa-sha2
```

What should happen is the error message should say words to the effect of:

```
keyword auth, invalid value: rsa-sha2, expected values: x,y,x,a,b,c
```

With operating systems deploying versions of libreswan with some features removed for all sorts of reasons, the man pages shipped with the package are probably inaccurate, online advice is probably stale, and anything AI related is a dead end as libreswan is smooshed with openswan is smooshed with freeswan is shmooshed with strongswan, the only option to debug this is to step through this in a debugger.

### Relevant log output

```shell

```

minfrin:
Further examples of why the possible values are so important, the next bit of the manual tells you to do this:

```
       leftautheap
           Whether the security gateways will authenticate uing an EAP method. Acceptable values are none (the default) and tls for EAPTLS. If EAP is the only authentication method, set leftauth=none in
           addition to leftautheap=tls=.
```

[...]

PR #2625, opened 2026-02-19
extract: list valid options in extract_enum_name() error messages

aamrindersingh:
extract_enum_name() just says the value is basically unrecognized without listing what is actually valid. The adjacent lookup_sparse_name() already does this via jam_sparse_names_quoted() add the equivalent jam_enum_names_quoted() for enum_names and wire it up.

Before: leftauth=rsa-sha2 invalid, 'rsa-sha2' unrecognized
After:  leftauth=rsa-sha2 is invalid, valid options are "RSASIG", "ECDSA", ..

Fixes: #2492

### DIFF
--- a/lib/libswan/enum_names.c
+++ b/lib/libswan/enum_names.c
@@ -375,9 +375,8 @@ static size_t jam_enum_name_quoted(struct jambuf *buf,
 				   long val)
 {
 	size_t s = 0;
-	name_buf nb;
 	s += jam_string(buf, "\"");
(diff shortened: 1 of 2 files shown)


aamrindersingh:
@cagney This addresses the /* include allowed names? */ TODO in extract_enum_name(). I followed the same jam_sparse_names_quoted() pattern from lookup_sparse_name(). 
lmk if any changes

aamrindersingh:
Done with both changes, ready for the test run.
Thanks

cagney:
I've got back test results; nice work.

PR #2627, opened 2026-02-19
ikev2: log digital signature algorithm name in IKE_AUTH request

### LINKED ISSUE #2490: in `sent IKE_AUTH request to 192.1.2.23:UDP/500 with digital-signature and ` name the digital signature
### What needs to be done

... if possible

aamrindersingh:
Solves #2490

The "sent IKE_AUTH request" log just says "digital-signature" when using RFC 7427. The signer and hash are already in st_v2_digsig at that point, so append them.

Before: sent IKE_AUTH request to ... with digital-signature and ...
After:  sent IKE_AUTH request to ... with digital-signature RSASSA-PSS with SHA2_512 and ...

Same format the "authenticated peer" log in keys.c already uses.


aamrindersingh:
@cagney This addresses the "name the digital signature" request in #2490. Have not run the tests yet , still setting up KVM locally.

aamrindersingh:
@cagney KVM setup is done. Updated around 257 test expected outputs to match the new log format , each one with the correct signer and hash for its cert type (RSASSA PSS/SHA2_512 for standard RSA, ECDSA/SHA2_384 for P 384, etc). Will push the test updates after running the full suite to verify.

cagney (changes requested):
The resulting output is exactly what is desired!
Just needs a few overly conservative checks for NULL.

(btw, can you please kill off the merge commit; it's a distraction)

cagney:
testing

PR #2630, opened 2026-02-23
confwrite: fix kt_sparse_name case never printing its value

mohanakatari119-bit:
The kt_sparse_name case in confwrite_value() was printing the key name followed by = but never printing the actual value. The loop found the matching sparse_name entry and broke out without calling fprintf for the name, leaving output like "type=" with no value and no newline.

Moved the fprintf inside the loop so it prints the full "key=value" line before breaking.


cagney (changes requested):

cagney:
I've tested 17aff54abf4be8ce148a65c081039778393f3315 with no problems with good tests.

> I'm curious as to how you found this.

Now I'm really curious :-)

Anyway, the WIP test readwriteconf-01 demonstrates the bug; I'm not sure why it is wip, but has been since it was added.  I'll fix that before merging.

PR #2633, opened 2026-02-24
pluto: allow --stderrlog to override logfile from config

### LINKED ISSUE #2599: `ipsec pluto --config /etc/ipsec.conf --stderrlog` can't override `logfile=/tmp/pluto.log`
### What needs to be done

Given a typical testsuite `ipsec.conf` that contains:
```
config setup
	# put the logs in /tmp for the UMLs, so that we can operate
	# without syslogd, which seems to break on UMLs
	logfile=/tmp/pluto.log
	logtime=no
	logappend=no
	plutodebug=all
	dumpdir=/tmp
	nhelpers=1
```
it isn't possible to both start pluto with that file and force logging to stderr.

- `--stderrlog` fails as `logfile=` is checked first
- `--stderrlog --logfile=` fails as `--logfile` (currently) requires a non-empty argument

Perhaps `--logfile=/dev/stderr` works.

Vinnu124:
initially --stderrorlog was not overriding the --logfile even if it was empty.

Fixed by making --stderrlog clear the logfile path before the file-open check in switch_log(). Also allowed --logfile= (empty) to explicitly unset the config's logfile by changing it from required_argument to optional_argument

fixes #2599

### DIFF
--- a/programs/pluto/log.c
+++ b/programs/pluto/log.c
@@ -81,6 +81,12 @@ void switch_log(struct logger **logger)
 	 * Hence save the new log file in LOG_FILE and switch at the
 	 * end.
 	 */
+	if (log_param.log_to_stderr) {
+		log_param.log_to_file = NULL;
+	}
+	if (log_param.log_to_file != NULL && log_param.log_to_file[0] == '\0') {
+		log_param.log_to_file = NULL;
+	}
 	PASSERT((*logger), pluto_log_file == stderr);
 	FILE *log_file = NULL;
 
--- a/programs/pluto/ipsec-pluto.8.xml
+++ b/programs/pluto/ipsec-pluto.8.xml
@@ -240,7 +240,10 @@
 	  </term>
           <listitem>
             <para>
-	      direct logging to standard error instead of a log file
+	      direct logging to standard error instead of a log file.
+	      When specified, overrides any
+	      <property>logfile=</property> setting from the
+	      configuration file, regardless of argument order.
 	    </para>
 	    <para>
 	      Often combined with <option>--nofork</option> debugging
@@ -257,6 +260,14 @@
 	      direct logging to
 	      <filename><replaceable>filename</replaceable></filename>
 	      instead of &syslog.3;
+	    </para>
+	    <para>
+	      When <replaceable>filename</replaceable> is empty
+	      (<option>--logfile=</option>), any
+	      <property>logfile=</property> setting from the
+	      configuration file is cleared, causing
+	      <command>pluto</command> to fall back to
+	      <option>--stderrlog</option> or &syslog.3;.
 	    </para>
 	    <para>
 	      See &ipsec.conf.5; and
--- a/programs/pluto/plutomain.c
+++ b/programs/pluto/plutomain.c
@@ -718,7 +718,7 @@ int main(int argc, char **argv)
 			continue;
 
 		case OPT_LOGFILE:	/* --logfile */
-			update_setup_string(KSF_LOGFILE, optarg_nonempty(logger));
+			update_setup_string(KSF_LOGFILE, optarg_empty(logger));
 			continue;
 
 		case OPT_DNSSEC_ENABLE:	/* --dnssec-enable yes|no */
--- a/testing/pluto/whack-pluto-02-logfile/description.txt
+++ b/testing/pluto/whack-pluto-02-logfile/description.txt
@@ -1,0 +1,8 @@
+test --logfile and --stderrlog command-line options
+
+- --logfile /tmp/test.log writes to a file
+- --logfile '' (empty) disables file logging, falls back to syslog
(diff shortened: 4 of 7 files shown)


cagney:
I need to think about this:

> changing it from required_argument to optional_argument

would mean that `--logfile /tmp/pluto.log` is no longer valid.

programs/pluto/log.c
```diff
@@ -81,6 +81,14 @@ void switch_log(struct logger **logger)
...
 	 * end.
 	 */
+	if (log_param.log_to_stderr)
+	{
+		log_param.log_to_file = NULL;
+	}
+	if (log_param.log_to_file != NULL && log_param.log_to_file[0] == '\0')
+	{
+		log_param.log_to_file = NULL;
+	}
```
cagney: this indentation style doesn't match the rest of the function
(It's roughly linux kernel style; in emacs it's c-set-style linux)

cagney:
> if the --stderrorlog is present , the above would not be valid , not sure why normally when set to optional it wont be valid.

It's in the fine print of getopt() (technically getopt_long()).  With optional_argument, getopt() expects `=`  as an indicator that there is an argument following.  Hence `--logfile=/tmp/pluto.log` is ok, but `--logfile /tmp/pluto.log` no longer works.

For instance, in `algparse` `--pfs` has an optional argument.  When there's no `=`, getopt() treated `no` as the parameter, contrast:
```
$ ./OBJ.linux.x86_64/programs/algparse/algparse --pfs=no ike=aes_gcm
algparse -v2 'ike=aes_gcm'
	AES_GCM_16-HMAC_SHA2_512+HMAC_SHA2_256-DH19+DH20+DH21+DH31+MODP4096+MODP3072+MODP2048+MODP8192
```
with
```
$ ./OBJ.linux.x86_64/programs/algparse/algparse --pfs no ike=aes_gcm
algparse -v2 -pfs 'ike=no'
	ERROR: IKE Encryption Algorithm (cipher) 'no' is not recognized
...
```
(algparse is easier to play with and hack than pluto)

Keeping required argument and then allowing logfile='' and --logfile '' might do the trick.  I'm not sure.  Thoughts?

(code actually uses optarg.[hc] but that is just wrapping getopt_long() up in a pretty bow)

Vinnu124:
This makes sense. 
Since in this PR , the --stderrflag is being checked first , keeping the argument as required and allowing --logfile '' seems like the perfect fix.
Will change the argument type back to required.

cagney:
Nice progress.

Because this is a functional change it should be added to the testsuite.   And since logfile= et.al. can be set in ipsec.conf, and then overriden by command line options, there are going to be some interesting combinations.

I'd suggest making a copy of `whack-pluto-01-startup` and using that as a starting point.  Even if you're not in a position to run this test, it should be possible to mock something up.

Vinnu124:
Added the Required Test Cases.

cagney:
Here's the output from a test run:
[west.console.diff.log](https://github.com/user-attachments/files/25640962/west.console.diff.log)
```
@@ -7,26 +7,27 @@
  # Test 1: --logfile <path>
 west #
  ipsec pluto --logfile /tmp/test.log --config /etc/ipsec.conf
+warning: ipsec pluto: /etc/ipsec.conf:6: overriding earlier 'config setup' keyword with new value: logfile=/tmp/pluto.log
 west #
```
the order matters, I'd also try `--config ... --logfile ...`.

```
 west #
  ipsec pluto --stderrlog --nofork --config /etc/ipsec.conf &
+[x] PID
 west #
```
when running stuff in the background libreswan has a testsuite convention of adding `sleep 1` as in:
```
  ipsec pluto --stderrlog --nofork --config /etc/ipsec.conf & sleep 1
```
it isn't exactly obvious but its to help prevent `ipsec pluto` output being intermingled with `[x] PID` from the shell

however ...

```
  ipsec pluto --stderrlog --nofork --config /etc/ipsec.conf &
+[x] PID
 west #
  ../../guestbin/wait-until-pluto-started
+Starting Pluto (Libreswan Version ...) pid:PID
+| debug: routing+base+cpu-usage+refcnt
```
Here, it might be better to use `ipsec pluto --self-check ...` and drop the `&`; oh and remove debug from ipsec.conf.

also, what about the reverse order?

cagney:
btw, can I suggest considering a re-base / rewrite so that this is reduced to two commits:

- the logging change
- the testsuite

Vinnu124:
Thanks for the feedback.
Updated the PR with --selftest (dropped the '&' and sleep) for running in background. added a few more test cases as well.

cagney:
the updated test passed! doing a full test run

cagney:
I've merged this, I esp like the test with `--logfile ''`
@Vinnu124, assuming you're a gsoc candidate, could you please look through libreswan's updated guidance

Vinnu124:
Sure will go through it.

PR #2634, opened 2026-02-24
pluto: split updown command into argv[] when updown-config=exec

### LINKED ISSUE #2628: when updown-config-exec, turn `ipsec _updown` into `argv[] = { "ipsec", "_updown", NULL, }`
### What needs to be done

ssia; the function `ttoshunks()` is likely useful here

stenaflow:
Fixes #2628 

When updown-config is in exec mode the command was passed to execve() as a single string ("ipsec _updown").

This patch tokenizes the updown command using ttoshunks() as suggested in the issue.

Tested with testing/pluto/basic-pluto-*

### DIFF
--- a/programs/pluto/updown.c
+++ b/programs/pluto/updown.c
@@ -131,10 +131,10 @@ static bool build_updown_exec(struct updown_exec *exec,
 				return false;
 		}
 
-		// convert the command string into a shrunk
+		// convert the command string into a shunk
 		shunk_t input = {
-				.ptr = exec->cmd,
-				.len = strlen(exec->cmd),
+			.ptr = exec->cmd,
+			.len = strlen(exec->cmd),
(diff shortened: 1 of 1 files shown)


cagney:
@stenaflow I have some general comments, I'll attach them to the bug.

cagney:
https://github.com/libreswan/libreswan/issues/2628#issuecomment-3960274905

stenaflow:
> @stenaflow I have some general comments, I'll attach them to the bug.

Thank you, I will go into that

cagney (changes requested):
please eliminate the two fixed-size buffers and fix the leak
could you please come up with a strategy for testing this

please also:
- rebase
- squash the changes into one
- review the diff; look for and remove any unnecessary changes

PR #2636, opened 2026-02-24
extract: wire up ignore-peer-dns to connection config

### LINKED ISSUE #2539: `--ignore-peer-dns` seems to go no where
### What needs to be done

nothing in extract.c to extract it

aamrindersingh:
Fixes #2539

The ignore-peer-dns option was parsed and passed through whack but the extract_bool() call was missing in extract_connection(), so config->ignore_peer_dns was never set, Added the missing extract_bool() call

also test that verifies ignore-peer-dns=yes causes DNS payloads to be ignored and resolv.conf to remain unchanged.


cagney (changes requested):
@aamrindersingh I did some digging, has been broken since 5.0, so you'll get a CHANGES entry.

Here's some notes:
- the testsuite and code changes are separate commits when possible
it's going to be --no-ff merged
- `ipsec auto --add` et.al. is old school; `ipsec add ...` et.al. is prefered
- on nat, can it use `../../guestbin/nic-nat.sh 192.1.3.0/24 192.1.2.254 50000`
or can WEST instead be used removing need for NIC

aamrindersingh:
@cagney 
Done, went with WEST to keep it simple as NIC is not needed for testing ignore peer dns. Also split code and test into separate commits and switched to ipsec add/up/down
Thanks

aamrindersingh:
Messed up the commits , all clean now 👍

cagney (changes requested):
Just some tweeks:

- when adding a test, it also needs to be added to `testing/pluto/TESTLIST` (something easily and often missed)
- can I suggest a rebase / rewrite
might as well combine the testsuite changes

I'll then do a full testrun

aamrindersingh:
@cagney Done, Thanks
You can go ahead with the test suite run, all good on my kvm setup

cagney:
@aamrindersingh could you please rebase; the way `extract_bool()` is called was changed.
```
<<<<<<< HEAD
	config->ikev2_pam_authorize =
		extract_bool(kv(wm, END_ROOF, KWS_PAM_AUTHORIZE),
			     /*value_when_unset*/YN_NO,
			     &d, verbose);
=======
	config->ikev2_pam_authorize = extract_bool("", "pam-authorize",
						   wm->wm_pam_authorize,
						   /*value_when_unset*/YN_NO,
						   wm, &d, verbose);
	config->ignore_peer_dns = extract_bool("", "ignore-peer-dns",
					       wm->wm_ignore_peer_dns,
					       /*value_when_unset*/YN_NO,
					       wm, &d, verbose);
>>>>>>> aamrindersingh/amrindersingh/2539-ignore-peer-dns-extraction
```

cagney (changes requested):
(useless github, I can't mark the PR as please update without fudging up a review of a specific patch)

aamrindersingh:
@cagney Rebased , Resolved, Pushed 👍

cagney (approved):
nice work

PR #2660, opened 2026-03-12
ikev2: log initiator's proposed traffic selectors

### LINKED ISSUE #2534: log the initiators TS somewhere
### What needs to be done

a bit like the ike= line
the initiator should log the TS proposal being sent, it could be:
- during connection orientation when TS proposal is built
- in the send line
- in the TS rejected line

OsemaFadhel:
Fixes #2534

Log the traffic selector proposed by the initiator.

Example output:

`"west-cuckold" #1: sent IKE_AUTH request to 192.1.2.23:UDP/500 with shared-key-mac and FQDN '@west'; Child SA #2 {ESP <0xESPESP} [192.0.3.0/24===192.0.2.0/24]`

`"west-cuckoo" #3: sent CREATE_CHILD_SA request to create Child SA using IKE SA #1 {ESP <0xESPESP} [192.0.3.0/24===192.0.20.0/24]
`


cagney:
testing

cagney (changes requested):
briefly

- ikev2-cp-02-ignore-peer-dns was missed
- addresspool-4in4-exhausted-ikev1-ikev2 and addresspool-4in4-shared-ikev1-ikev2 don't need tweaking, the responder was slow
- impair-14-ike-responder-spi-0-ikev2 shouldn't need tweaking
the reponder was really slow

I'll follow up with the latter.

PR #2670, opened 2026-03-14
Solution for GitHub Task #2652

### LINKED ISSUE #2652: add, and then use, `str_said_sensitive()`
### What needs to be done

The include/ip*.h code has two versions of most to-string functions.  For instance, there's the pair:
```
size_t jam_address(struct jambuf *buf, const ip_address *src);
const char *str_address(const ip_address *src, address_buf *dst);
```
and the pair:
```
const char *str_address_sensitive(const ip_address *src, address_buf *dst);
size_t jam_address_sensitive(struct jambuf *buf, const ip_address *src);
```
 `ip_said` is missing the second set (watch out for ipcheck).

--

`kernel_ops.c` contains two calls to `str_said()`, with above in place there's a cleanup opportunity.

shahrinf:
added str_said_sensitive() and jam_said_sensitive() to lib/include/ip_said.[h/c]
they hide the ip based on log_ip
then modified kernel_ops_del_ipsec_spi() in programs/pluto/kernel_ops.c, making use of str_said() and str_said_sensitive()

### DIFF
--- a/include/ip_said.h
+++ b/include/ip_said.h
@@ -109,6 +109,8 @@ typedef struct {
 
 size_t jam_said(struct jambuf *buf, const ip_said *said);
 const char *str_said(const ip_said *said, said_buf *buf);
+size_t jam_said_sensitive(struct jambuf *buf, const ip_said *said);
+const char *str_said_sensitive(const ip_said *said, said_buf *buf);
 
 /*
  * Details.
--- a/lib/libswan/ip_said.c
+++ b/lib/libswan/ip_said.c
@@ -93,11 +93,29 @@ size_t jam_said(struct jambuf *buf, const ip_said *said)
 	return s;
 }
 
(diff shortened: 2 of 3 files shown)


cagney:
Screenshot From 2026-03-16 17-41-56

cagney (changes requested):
The code looks fine, however, the editor you're using isn't configured right and is mangling white space.

I've attached a screenshot showing part of what I see when I look at the files with emacs `(setq-default show-trailing-whitespace t)` - there should be no trailing spaces and spaces not tabs.

cagney:
ps you should squash the second commit into the first

(if you want, you could also split the changes to include & lib adding the new functions; and the change to programs that uses it; I'm not fussed)

shahrinf:
@cagney  
thanks for the review
i'll fix those things

cagney:
testing for #2652

cagney:
nice cleanup

Issue #2556, opened 2025-12-18
dns hostnames never looked up again during revive


wofferl:
### Please Tell Us About Your Operating System

Libreswan 5.3

### Please Describe Your IPsec Configuration

Connections with dynamic host names like`right=dyndns.local`

### Please Tell Us What Happened?

Since 5.x when using something like `right=dyndns.local` the hostname is never looked up again. The ddns code only look up names again that could not resolved at startup.

I have changed this for 5.3 where the lookup is done every time before initiating a connection with KH_IPHOSTNAME and I would create a PR if you think this change is useful. 
But now seeing heavy changes to this codes in 5.4, the question is if you already working on this or what plans are actual existing for the dns code.
My goal would be to look up on initiating as I already have done, and change the existing ddns code to also refresh already looked up hostnames for passive connections.

Here an example for the actual different behaviour between 4.15 and 5.3

In 4.15 this could be done with the `dpdaction=restart_by_peer`:
```
#1: STATE_V2_ESTABLISHED_IKE_SA: retransmission; will wait 32 seconds for response
#1: ERROR: asynchronous network error report on ens19 (10.0.11.64:500), complainant 10.0.12.103: Connection refused [errno 111, origin ICMP type 3 co
de 3 (not authenticated)]
#1: STATE_V2_ESTABLISHED_IKE_SA: 60 second timeout exceeded after 7 retransmits.  No response (or no acceptable response) to our IKEv2 message
#1: liveness action - restarting all connections that share this peer
terminating SAs using this connection
#1: deleting state (STATE_V2_ESTABLISHED_IKE_SA) aged 94.105536s and sending notification
#2: ESP traffic information: in=0B out=0B
#3: initiating IKEv2 connection
#3: sent IKE_SA_INIT request to 10.0.12.102:500
```
With 5.3 the revive code immediately starts without lookup the ip address. 
```
#1: retransmitting liveness probe informational request; will wait 32 seconds for response
#1: ERROR: asynchronous network error report on ens19 (10.0.12.103:500), complainant 10.0.11.64: Connection refused [errno 111, origin ICMP type 3 co
de 3 (not authenticated)]
#1: ESTABLISHED_IKE_SA: 60 second timeout exceeded after 7 retransmits.  No response (or no acceptable response) to our IKEv2 message
#2: connection is supposed to remain up; revival attempt 1 scheduled in 0 seconds
#2: ESP traffic information: in=0B out=0B
[...]

PR #2687, opened 2026-03-17
2556 ddns revival

cagney:
#2556


cagney:
```
| event CONNECTION_CHECK_DDNS for "named" in 60s: DDNS:  starting check_dns
| no whack to attach
| DDNS:   skipping connection "named", already has addresses
```

cagney:
what about running DNS always, and then leaving it to failureshunt= to decide of things should be unrouted

cagney:
west doesn't realize that new DNS is on a new interface
```
| sending 732 bytes for IKE_SA_INIT request through eth1 from 192.1.2.45:500 to 192.0.1.15:500 using UDP (for #4)
```

PR #2701, opened 2026-03-22
kernel_xfrm: remove description param from sendrecv_xfrm_msg()

### LINKED ISSUE #2653: drop `description` parameter to `sendrecv_xfrm_msg()`?
### What needs to be done

It's thought to be redundant.

sranish:
fixes #2653 

**Summary of changes**

Based on suggestion that description is redundant, kernel_xfrm: drop redundant description parameter from sendrecv_xfrm_msg().

The description parameter is redundant as the story parameter already provides sufficient context for logging and diagnostics.

Remove the description parameter from sendrecv_xfrm_msg() and update all callers accordingly. Simplify logging statements to rely solely on story param.

No functional change intended.

### DIFF
--- a/programs/pluto/kernel_xfrm.c
+++ b/programs/pluto/kernel_xfrm.c
@@ -505,7 +505,9 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 {
 	size_t len = hdr->nlmsg_len;
 
-	ldbg(logger, "%s() sending %d %s", __func__, hdr->nlmsg_type, story);
+	name_buf sb;
+	ldbg(logger, "%s() sending %s %s", __func__,
+		str_sparse_long(&xfrm_type_names, hdr->nlmsg_type, &sb), story);
 	if (LDBGP(DBG_TMI, logger)) {
 		LDBG_dump(logger, hdr, len);
 	}
--- a/programs/pluto/kernel_xfrm.c
+++ b/programs/pluto/kernel_xfrm.c
@@ -120,7 +120,7 @@ static void netlink_process_xfrm_messages(struct verbose verbose, int fd, void *
 static void netlink_process_rtm_messages(struct verbose verbose, int fd, void *arg);
 static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
 			      unsigned expected_resp_type, struct nlm_resp *rbuf,
-			      const char *description, const char *story,
+			      const char *story,
 			      int *recv_errno,
 			      const struct logger *logger);
 static err_t xfrm_iptfs_ipsec_sa_is_enabled(struct logger *logger);
@@ -437,7 +437,7 @@ static void kernel_xfrm_flush(struct logger *logger)
 		.nlmsg_len = NLMSG_ALIGN(NLMSG_LENGTH(0)),
 	};
 	sendrecv_xfrm_msg(&policy, NLMSG_ERROR, &rsp,
-			  "flush", "policy",
+			  "policy",
 			  &recv_errno, logger);
 
 	struct {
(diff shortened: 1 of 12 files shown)


cagney (changes requested):

cagney:
While it shouldn't affect output; is this known for certain?

sranish:
> While it shouldn't affect output; is this known for certain?

well, description was redundant with story at the call sites. I dont think dropping description should lose actionable context here.
 
If this turns out to regress readability in some paths, it should be trivial to fold the verb into story at those call sites or add the symbolic msg type to that specific log as well.

cagney (changes requested):
This change, it turns out, is going to affect the testsuite.  My test run is throwing up this difference:
```
-ERROR: "westnet-eastnet-second" #3: netlink response for Del SA esp.ESPSPIi@192.1.2.23: No such process (errno 3)
+ERROR: "westnet-eastnet-second" #3: netlink response for esp.ESPSPIi@192.1.2.23: No such process (errno 3)
```
once the name string is added to the error, it will look more like (yes I also added "of"):
```
+ERROR: "westnet-eastnet-second" #3: netlink response for XFRM_MSG_DELSA of esp.ESPSPIi@192.1.2.23: No such process (errno 3)
```
sed is a useful tool here.

cagney:
@ANISH-SR ping

sranish:
> This change, it turns out, is going to affect the testsuite. My test run is throwing up this difference:
> 
> ```
> -ERROR: "westnet-eastnet-second" #3: netlink response for Del SA esp.ESPSPIi@192.1.2.23: No such process (errno 3)
> +ERROR: "westnet-eastnet-second" #3: netlink response for esp.ESPSPIi@192.1.2.23: No such process (errno 3)
> ```
> 
> once the name string is added to the error, it will look more like (yes I also added "of"):
> 
> ```
> +ERROR: "westnet-eastnet-second" #3: netlink response for XFRM_MSG_DELSA of esp.ESPSPIi@192.1.2.23: No such process (errno 3)
> ```
> 
> sed is a useful tool here.

Hi @cagney  , I’ve pushed a fix to address the testsuite diff. Would really appreciate your review when you get a chance.

programs/pluto/kernel_xfrm.c
```diff
@@ -622,16 +626,25 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
...
 			if (LDBGP(DBG_BASE, logger)) {
-				LDBG_log(logger, "%s() netlink response for %s included non-error error",
-					 __func__, story);
+				name_buf sb;
+				LDBG_log(logger, "%s() netlink response for %s of %s included non-error error (%s)",
+					__func__,
+					str_sparse_long(&xfrm_type_names, hdr->nlmsg_type, &sb),
+					story,
+					str_sparse_long(&xfrm_type_names, rsp.n.nlmsg_type, &sb));
 				llog_ext_ack(DEBUG_STREAM, logger, &rsp.n);
```
cagney, line 635: Good catch, there's two messages involved.

programs/pluto/kernel_xfrm.c
```diff
@@ -622,16 +626,25 @@ static bool sendrecv_xfrm_msg(struct nlmsghdr *hdr,
...
+			name_buf sb;
+			const char *msg_type = str_sparse_long(&xfrm_type_names, hdr->nlmsg_type, &sb);
+			/* Convert XFRM_MSG_DELSA to "Del SA" for test compatibility */
+			const char *display_type = (hdr->nlmsg_type == XFRM_MSG_DELSA) ? "Del SA" : msg_type;
 			llog_errno((streq(story, "Probe Test") ? ALL_STREAMS : ERROR_STREAM),
 				   logger, -rsp.u.e.error,
-				   "netlink response for %s: ",
+				   "netlink response for %s %s: ",
+				   display_type,
 				   story);
```
cagney, line 648: Just FYI. Instead of all the temp variables, the message can also be constructed using LLOG_JAMBUF() but that's getting pretty obscure.

cagney (approved):
Looks ok.
Expect a rebase to be pushed before the merge.

cagney:
I needed to tweak some test output

Issue #2503, opened 2025-10-15
add auth=digsig, or similar


cagney:
### What needs to be done

instead of specifying `auth={rsasig,ecdsa,eddsa,...}` specify `auth=digcig` and then let the certificate and/or private key determine the details.

This growing list, for instance, could be eliminated:
```
	case IKEv2_AUTH_DIGITAL_SIGNATURE:
		return LELEM(AUTH_RSASIG) | LELEM(AUTH_ECDSA);
```

cagney:
@ruiliio I've changed this to a feature. 

There's a messy interaction between {left,right}auth= and authby=.

Given `auth=digsig` does this mean that from https://www.iana.org/assignments/ikev2-parameters/ikev2-parameters.xhtml#ikev2-parameters-12 only:
```
14 	Digital Signature 	[[RFC7427](https://www.iana.org/go/rfc7427)]
```
is allowed for the AUTH payload?  Which exact algorithm being determined by authby=, which defaults to:
```
#define AUTHBY_IKEv2_DEFAULTS (struct authby) { .rsasig = true, .rsasig_v1_5 = true, .ecdsa = true, }
```

cagney:
With auth=digsig, the responder, during IKE_SA_INIT, will need to find connection peer's acceptable cred type it can send back.  Except, it may have the wrong connection.

ruiliio:
Sorry for the late reply. In my current PR #2708, 'auth=digsig' uses method 14 when peer supports RFC 7427, and falls back to legacy methods (1 for RSA, 9/10/11 for ECDSA) when peer doesn't. So not strictly method 14 only. 

The algorithm is narrowed by authby=  via intersection. If authby= is not explicitly set, it defaults to AUTHBY_DIGITAL_SIGNATURE since AUTHBY_IKEv2_DEFAULTS omits eddsa.

cagney:
having worked through some of the code a problem with `digsig` is it is somewhat loaded - it could refer to any digital signature algorithm or just the digital signature algorithms supported by the digital signature auth payload, or even just a subset of that.

Suggest instead calling it `authby=pubkey` since that, within pluto refers to any digsig like algorithm.
This also means that it should include all allowed algorithms for the ike version.

PR #2708, opened 2026-03-24
Add auth=digsig to determine signature algorithm from certificate at runtime

ruiliio:
Fix/Task #2503

Add auth=digsig option in ipsec.conf (leftauth/rightauth=digsig) as an umbrella type  for all digital signature methods. The cert's key type determines the signature algorithm at runtime. Changes include: 
  - add get_cert_key_type() helper to get key type from local certificate
  - return null in get_local_private_key for cert/key type mismatch to enable multi type private keys probing
  - backwards compatible: legacy fallbacks when peer doesn't send N(SIGNATURE_HASH_ALGORITHMS)
  - peeks at AlgorithmIdentifier length in IKEv2_AUTH_DIGITAL_SIGNATURE payload (proposed_v2AUTH()) to determine concrete auth type
  - check concrete auth proposal against authbys in peer_id.c for less strict matching process
  
Testing:
  - update authbycheck for AUTH_DIGSIG
  - add 10 auth=digsig KVM tests covering the code changes
  - add 2 ECDSA multicert road-warrior KVM tests to validate the less strict matching in peer_id.c
  - tested with 77 related tests for regression (including 6 strongSwan interop tests)


cagney (changes requested):
I appreciate you making this a separate commit, and keeping library changes separate.  Unfortunately it doesn't compile, for instance:
```
ikev2_auth.c: In function ‘local_v2AUTH_method’:
ikev2_auth.c:169:9: error: enumeration value ‘AUTH_DIGSIG’ not handled in switch [-Werror=switch]
  169 |         switch (authby) {
```
this just needs patching up.

I suspect you also need to pull the change to authbycheck as I see both `check-01` and `check-02-enumcheck` failing.

programs/pluto/extract.c
```diff
@@ -1746,15 +1746,15 @@ static diag_t extract_host_end(enum end end,
...
+	case AUTH_EDDSA:
+#ifndef USE_EDDSA
+		return diag("auth=eddsa is not supported by this build");
+#endif
 	case AUTH_RSASIG:
 	case AUTH_ECDSA:
-	case AUTH_EDDSA:
-#ifdef USE_EDDSA
+	case AUTH_DIGSIG:
 		authby_mask = authby_from_auth(auth);
```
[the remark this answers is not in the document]
ruiliio: Thank you!

programs/pluto/ikev2_auth.c
```diff
@@ -131,6 +131,29 @@ struct crypt_mac v2_calculate_sighash(const struct ike_sa *ike,
...
 
+static const struct pubkey_signer *digsig_select_signer(struct ike_sa *ike)
+{
+	const struct connection *c = ike->sa.st_connection;
+	struct authby authby = c->local->host.config->authby;
+	const struct pubkey_signer *signer = NULL;
+
+	if (authby.rsasig &&
+	    get_local_private_key(c, &pubkey_type_rsa, ike->sa.logger) != NULL) {
+		signer = &pubkey_signer_digsig_rsassa_pss;
```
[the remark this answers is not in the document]
ruiliio: Thanks for the review. Replaced the key probing with get_cert_key_type() to get the key type from cert and then check it against authby policy.

ruiliio:
Patched up the lib and pluto changes into one commit and reworked the helper logic.

cagney:
@ruiliio I'll be looking at this.  But, to be up front, I'm not sure when it will be merged:

- there's this mythical beast call 5.4 threatening to appear, I think this should go after that drops

- this is going to interact with the GSoC project for publishing auth

ruiliio:
I understand. No rush. Let me know if anything needs to be changed or rebased.

lib/libswan/auth_names.c
```diff
@@ -24,6 +24,7 @@ static const char *const auth_name[] = {
 	S(AUTH_PSK),
 	S(AUTH_RSASIG),
 	S(AUTH_ECDSA),
+	S(AUTH_DIGSIG),
```
cagney, line 27: the test `check-02-enumcheck` will need an update

configs/d.ipsec.conf/conn/auth.xml
```diff
@@ -16,7 +16,10 @@
       based authentication with SHA2-384, <option>rsa-sha2_512</option>
       for RSA-PSS digital signatures based authentication with SHA2-512,
       <option>ecdsa</option> for ECDSA digital signatures based
-      authentication, <option>secret</option> for shared
+      authentication, <option>digsig</option> to let the
+      certificate determine the digital signature algorithm at
+      runtime (RSA-PSS, ECDSA, or EdDSA),
+      <option>secret</option> for shared
       secrets (PSK) authentication and <option>null</option> for
```
cagney, line 23: suggest:
..., `digsig`for for RFC7427 based digital signature authentication (the certificate will determine the digital signature algorithm during negotiation), ...

programs/pluto/ikev2_auth.c
```diff
@@ -131,6 +132,27 @@ struct crypt_mac v2_calculate_sighash(const struct ike_sa *ike,
...
+	SECKEYPublicKey *pubkey = CERT_ExtractPublicKey(nss_cert);
+	if (pubkey == NULL) {
+		ldbg(ike->sa.logger, "cannot extract public key from certificate");
+		return NULL;
+	}
+
+	const struct pubkey_type *type = pubkey_type_from_SECKEYPublicKey(pubkey);
+	SECKEY_DestroyPublicKey(pubkey);
+	return type;
+}
```
cagney: Yes!  If annoying.  cert_t doesn't know its type.

Just move the NSS code to `secrets.h` and `certs.c` so that  `ikev2_auth.c` doesn't need to know these internals.

lib/libswan/authby.c
```diff
@@ -114,6 +114,7 @@ struct authby authby_from_auth(enum auth auth)
 	case AUTH_ECDSA: return (struct authby) { .ecdsa = true, };
 	case AUTH_EDDSA: return (struct authby) { .eddsa = true, };
 	case AUTH_RSASIG: return (struct authby) { .rsasig = true, .rsasig_v1_5 = true };
+	case AUTH_DIGSIG: return AUTHBY_DIGITAL_SIGNATURE;
```
cagney, line 117: Does DIGSIG need to be added to AUTHBY?  I'll explain why in a separate note ...

programs/pluto/ikev2_auth.c
```diff
@@ -131,6 +132,27 @@ struct crypt_mac v2_calculate_sighash(const struct ike_sa *ike,
 	return crypt_hash_hunks("sighash", hasher, &blobs.hunks, ike->sa.logger);
 }
 
+/* Get the key type of the local certificate, returns NULL on failure. */
+static const struct pubkey_type *get_cert_key_type(struct ike_sa *ike)
+{
+	const struct connection *c = ike->sa.st_connection;
+	CERTCertificate *nss_cert = c->local->host.config->cert.nss_cert;
```
cagney, line 139: That's a nice fix to an annoying gap - a cert_t should know its pubkey type.

Everything below belongs in `secrets.h` + `certs.c`, something like:
```
get_cert_key_type(cert_t, logger)
```
so that `ikev2_auth.c` doesn't gain another direct dependency on NSS.

programs/pluto/ikev2_auth.c
```diff
@@ -258,6 +280,51 @@ enum ikev2_auth_method local_v2AUTH_method(struct ike_sa *ike,
...
+		 * Peer sent us N(SIGNATURE_HASH_ALGORITHMS)
+		 * indicating a preference for Digital Signature
+		 * Method, and local policy was ok with the
+		 * suggestion.
+		 */
+		pexpect(auth_in_authby(AUTH_DIGSIG, c->local->host.config->authby));
+		if (ike->sa.st_v2_digsig.negotiated_hashes != LEMPTY) {
+			return IKEv2_AUTH_DIGITAL_SIGNATURE;
+		}
+
```
cagney, line 294: Instead of expecting AUTH_DIGSIG in authby; can the code call `get_cert_key_type()` and check that the returned type is in authby?

Picking up on another comment, that hopefully means that authby= doesn't need digsig added.

This should also mean that the legacy fallbacks can be dropped.  auth=digsig implies rfc7427 only.

programs/pluto/ikev2_auth.c
```diff
@@ -872,6 +960,21 @@ stf_status submit_v2AUTH_generate_initiator_signature(struct ike_sa *ike,
 		case AUTH_EDDSA:
 			signer = &pubkey_signer_digsig_eddsa_ed25519;
 			break;
+		case AUTH_DIGSIG: {
+			const struct connection *c = ike->sa.st_connection;
+			const struct pubkey_type *type = get_cert_key_type(ike);
+			if (type == &pubkey_type_rsa && c->local->host.config->authby.rsasig) {
```
cagney, line 966: yep!

programs/pluto/peer_id.c
```diff
@@ -346,17 +346,19 @@ static bool score_host_connection(const struct ike_sa *ike,
...
-			vdbg("skipping because mismatched authby");
-			return false;
+		for (enum auth auth = AUTH_FLOOR; auth < AUTH_ROOF; auth++) {
+			if (LHAS(proposed_authbys, auth)) {
+				if (!auth_in_authby(auth, d->remote->host.config->authby)) {
+					vdbg("skipping because mismatched authby");
+					return false;
+				}
+				break;
+			}
```
cagney, line 359: Another code pain point.  `proposed_authbys` is a bit-set (lset_t) of `enum auth`.  Hence the need for the loop.

I've created #2780 for a follow-up.

cagney (commented):
I think  auth=digsig should be limited to rfc7427 methods.

Yes, we've code that tries to make auth=rsa... work with both the old and new RSA hash methods, but that is from a few years ago.  By this point no one should be using PKCS#1 1.5 RSA so lets not encourage it :-)

Hence, I'm wondering if the `authby` changes are needed.  auth=digsig would only imply a subset of the authby bits.

(As an aside this patch is far easier to read than the first draft)

lib/libswan/authby.c
cagney: There's been a bit of an IRC discussion about this.

Perhaps it should so that there's slightly more 1:1 between auth and authby, but it's a mess.
Thoughts?

cagney:
@ruiliio can I suggest sitting on your hands for now; this pull request has sparked a wider discussion.

lib/libswan/authby.c
ruiliio: Yes, I think `AUTH_DIGSIG` should return `{.rsasig, .ecdsa, .eddsa}`, RFC 7427 only, no `rsasig_v1_5`. Otherwise `digital_signature_in_authby(authby_from_auth(that_auth))` at `ikev2_auth.c:559` trips for any `rightauth=digsig` responder receiving wire 14.

programs/pluto/ikev2_auth.c
ruiliio: Calling `get_cert_key_type()` and checking the returned type here works. Even so, `authby` still needs digsig: `digital_signature_in_authby(authby_from_auth(that_auth))` at `:559` still trips for a `rightauth=digsig` responder on wire 14. And since `local_v2AUTH_method` is also called from `emit_local_v2AUTH`, the cert lookup would run twice per IKE_AUTH.  

Will drop the legacy fallback when this lands.

ruiliio:
Sounds good, let me know when it's good to continue. Thanks for the review!

cagney:
@ruiliio, @OsemaFadhel is going to extend `struct authby` so it contains a bit for all the cases.
Once that is in place this can be updated and merged.

cagney:
Grr, GitHub, Grr.  I can't note that this pull request is waiting on #2840

cagney:
based on this #2931 discussion, I think its possible to rebase and merge the changes to extract.c et.al. that map auth=digsig (and authby=digsig ?) onto a set of bits in the authby struct.

The current code in extract.c will then map one of the authby bits back to an arbitrary auth= choice (probably rsa).
Fixing this is a separate problem.  It might even be useful to, when auth=digsig, force it to UNSET, so code doesn't have to handle digsig.

This should leave the changes to ikev2_auth.c.  Again going on #2931, I think we'll need some groundwork - eliminate the internal auth config field (moved to struct state?) and a careful look at what @OsemaFadhel's branch contains.
-

PR #2741, opened 2026-04-12
fix #2713: add loglimit configuration option to allow disabling log rate limiting

### LINKED ISSUE #2713: `config setup loglimit=no`
### What needs to be done

to disable log limiting; a more advanced syntax is:
```
loglimit=md,certificate
```
flags.[hc] is probably useful

chmodshubham:
- fix #2713: add loglimit configuration option to allow disabling log rate limiting.
- add man documentation for loglimit

Signed-off-by: Shubham Kumar [chmodshubham@gmail.com](mailto:chmodshubham@gmail.com)

### DIFF
--- a/configs/d.ipsec.conf/setup/loglimit.xml
+++ b/configs/d.ipsec.conf/setup/loglimit.xml
@@ -1,0 +1,25 @@
+<varlistentry id='setup.loglimit'>
+  <term>
+    <option>loglimit=&yn_option;</option>
+  </term>
+  <listitem>
+    <para>
+      Whether pluto's internal log rate limiting is enabled.  When
+      enabled, pluto limits the number of repetitive log messages
+      (such as malformed packets or bad certificates) to prevent
+      excessive logging under heavy load.  Setting this to
+      <option>&no;</option> disables all rate limiters, allowing every
+      log message to be emitted.  This is useful for debugging or
+      when full log visibility is required.
+    </para>
+    <para>
+      The rate limit counters are reset once every hour.
+    </para>
+    <para>
+      The default is &yes;.
+    </para>
+    <para>
+      This option was added in Libreswan 5.4.
+    </para>
+  </listitem>
+</varlistentry>
--- a/programs/pluto/log_limiter.c
+++ b/programs/pluto/log_limiter.c
@@ -15,12 +15,14 @@
  */
 
 #include <pthread.h>    /* Must be the first include file; XXX: why? */
+#include <limits.h>     /* for UINT_MAX */
 
 #include "log_limiter.h"
 #include "log.h"
 
 #include "defs.h"
 #include "demux.h"
+#include "ipsecconf/setup.h"
 
 static global_timer_cb reset_log_limiter;	/* type check */
 
@@ -31,6 +33,7 @@ struct limiter {
 	const unsigned limit;
 	volatile unsigned count;
 	const char *what;
+	bool unlimited;
 };
 
 struct limiter log_limiters[LOG_LIMITER_ROOF] = {
@@ -58,6 +61,9 @@ struct limiter log_limiters[LOG_LIMITER_ROOF] = {
 
 static unsigned log_limit(const struct limiter *limiter)
 {
+	if (limiter->unlimited) {
+		return UINT_MAX;
+	}
 	if (impair.log_rate_limit.enabled) {
 		return impair.log_rate_limit.value;
 	}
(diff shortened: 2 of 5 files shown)


programs/pluto/log_limiter.c
```diff
@@ -155,4 +165,14 @@ void init_log_limiter(struct logger *logger)
...
+
+	// loglimit=no disables all rate limiters 
+	if (!config_setup_yn(KYN_LOGLIMIT)) {
+		FOR_EACH_ELEMENT(limiter, log_limiters) {
+			limiter->disabled = true;
+		}
+		llog(RC_LOG, logger, "log rate limiting [disabled]");
+	} else {
+		llog(RC_LOG, logger, "log rate limiting [enabled]");
+	}
```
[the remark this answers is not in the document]
cagney: Copilot is an idiot.

It's suggesting a convoluting micro-optimization.  Better to keep the code parsing options separate to the code acting on them.

programs/pluto/log_limiter.c
```diff
@@ -70,6 +76,10 @@ enum stream log_limiter_stream(struct logger *logger, enum log_limiter log_limit
 {
 	struct limiter *limiter = &log_limiters[log_limiter];
 
+	if (limiter->disabled) {
+		return RC_LOG;
+	}
+
 	/* allow imparing to override specified limit */
```
cagney, line 79: this strikes me as redundant:

- log_limit() returns UINT_MAX; making it unlimited, not disabled
- log_limiter_stream() returns RC_LOG; disabling it, not unlimited

Digging a little, I now see there's also the call:
```
			if (limiter->count > log_limit(limiter)) {
				vlog("%s rate limited log reset",
				     limiter->what);
			}
```
which explains the need for the `log_limit()` tweak, and, to me suggests it as preferable; it also points to the field being called .unlimited.

programs/pluto/log_limiter.c
```diff
@@ -15,12 +15,14 @@
  */
 
 #include <pthread.h>    /* Must be the first include file; XXX: why? */
+#include <limits.h>
```
cagney: copilot is an idiot.

Exactly like you've done, libreswan always puts system includes at the top, and adds them explicitly.

What I do suggest is appending `/* for UINT_MAX */ to the line - it leaves a bread crumb for our future selves so we know why the include was added.

cagney (changes requested):

configs/d.ipsec.conf/setup/loglimit.xml
```diff
@@ -0,0 +1,19 @@
...
+  <listitem>
+    <para>
+      Whether pluto's internal log rate limiting is enabled.  When
+      enabled, pluto limits the number of repetitive log messages
+      (such as malformed packets or bad certificates) to prevent
+      excessive logging under heavy load.  Setting this to
+      <option>&no;</option> disables all rate limiters, allowing every
+      log message to be emitted.  This is useful for debugging or
+      when full log visibility is required.
+    </para>
```
cagney, line 14: Can I suggest moving the documentation to a separate commit.  It's common for the documentation and code to be reviewed separately (esp. in large projects which admittedly this isn't), having them separate is a good habit to get into.

Just `&no;`?  You nailed the format, and a separate paragraph stating the default.  Just add a paragraph that this option was added in 5.4.

I think it should mention that the limit is reset once every NNN-hours.

cagney (changes requested):

programs/pluto/log_limiter.c
chmodshubham: > copilot is an idiot.

Yes, I thought so. That’s why I disabled it from auto review. :)



> What I do suggest is appending `/* for UINT_MAX */ to the line - it leaves a bread crumb for our future selves so we know why the include was added.

sure

configs/d.ipsec.conf/setup/loglimit.xml
chmodshubham: > Can I suggest moving the documentation to a separate commit.

sure



> Just `&no;`

Yes, I initially thought to add multiple options like md, certificate, etc., but then in the issue itself, you suggested that loglimit=no will be sufficient (ref: https://github.com/libreswan/libreswan/issues/2713#issuecomment-4184090651), so I just keep it this way.

chmodshubham: > Just add a paragraph that this option was added in 5.4.
> 
> I think it should mention that the limit is reset once every NNN-hours.

sure

programs/pluto/log_limiter.c
chmodshubham: > this strikes me as redundant:
> 
> * log_limit() returns UINT_MAX; making it unlimited, not disabled
> * log_limiter_stream() returns RC_LOG; disabling it, not unlimited
> 
> Digging a little, I now see there's also the call:
> 
> ```
> 			if (limiter->count > log_limit(limiter)) {
> 				vlog("%s rate limited log reset",
> 				     limiter->what);
> 			}
> ```
> 
> which explains the need for the `log_limit()` tweak, and, to me suggests it as preferable; it also points to the field being called .unlimited.

hm, I agreed.

PR #2759, opened 2026-04-25
pluto: log default proposals at startup 

### LINKED ISSUE #2753: dump default proposals during startup
### What needs to be done

so that there's a one-stop-shop for algorithms and proposals

also consider changing it to debug-log when debugging is enabled

Vinnu124:
added a patch to print the default proposals on startup.
Fixes: #2753 

image

### DIFF
--- a/lib/libswan/ike_alg.c
+++ b/lib/libswan/ike_alg.c
@@ -1425,3 +1425,48 @@ void init_ike_alg(struct logger *logger)
 	 */
 	log_ike_algs(logger);
 }
+/*
+ * Shared protocol table for IKE, ESP, and AH.
+ * Pluto can override alg_is_ok
+ * in the proposal_policy with the real kernel_alg_is_ok.
+ */
+
+static bool default_alg_is_ok(const struct ike_alg *alg,
+			      const struct logger *logger)
+{
+	if (alg->type == &ike_alg_kem) {
+		/* require an in-process/ike implementation of DH */
+		return ike_alg_is_ike(alg, logger);
+	} else {
+		/* no kernel to ask! */
+		return true;
+	}
+}
+
+static const struct ike_alg_protocol ike_alg_ike_protocol = {
+	.name = "ike",
+	.parser = ike_proposal_parser,
+	.pfs_vs_dh = false,
+	.alg_is_ok = ike_alg_is_ike,
+};
+
+static const struct ike_alg_protocol ike_alg_esp_protocol = {
+	.name = "esp",
+	.parser = esp_proposal_parser,
+	.pfs_vs_dh = true,
+	.alg_is_ok = default_alg_is_ok,
+};
+
+static const struct ike_alg_protocol ike_alg_ah_protocol = {
+	.name = "ah",
+	.parser = ah_proposal_parser,
+	.pfs_vs_dh = true,
+	.alg_is_ok = default_alg_is_ok,
+};
+
+const struct ike_alg_protocol *ike_alg_protocols[] = {
+	&ike_alg_ike_protocol,
+	&ike_alg_esp_protocol,
+	&ike_alg_ah_protocol,
+	NULL,
+};
--- a/programs/pluto/plutomain.c
+++ b/programs/pluto/plutomain.c
@@ -86,6 +86,7 @@
 #include "ddos.h"
 #include "helper.h"
 #include "resolve_helper.h"	/* for init_resolve_helper() */
+#include "kernel_alg.h"	
 
 #ifndef IPSECDIR
 #define IPSECDIR "/etc/ipsec.d"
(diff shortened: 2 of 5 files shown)


cagney:
Thanks for working on this.

Two + One things:

- it should be a proper log
  it's intended for end-users, this way we can point at the startup logs when asked what the defaults are (they vary across systems); you may have noticed that the docs already say to do this
- can you make it closer to `algparse --v2 ike`:
```
Default IKE v2 Proposals:
	AES_GCM_16_256
	AES_GCM_16_128
	CHACHA20_POLY1305
	AES_CBC_256-HMAC_SHA2_512_256+HMAC_SHA2_256_128
	AES_CBC_128-HMAC_SHA2_512_256+HMAC_SHA2_256_128
```
so it is easier for end-users to read (I do like your inclusion of commas)
- this will affect some of the check-* tests, so the'll need updating

I'll also add some backstory to the bug.

cagney (changes requested):
see comment

programs/pluto/plutomain.c
```diff
@@ -586,6 +587,60 @@ static deltatime_t check_config_deltatime(enum config_setup_keyword kw,
...
+		const struct proposal_policy policy = {
+			.version = protos[i].version,
+			.alg_is_ok = protos[i].alg_is_ok,
+			.pfs = true,
+			.check_pfs_vs_ke = false,
+			.logger = logger,
+			.stream = DEBUG_STREAM,
+			.ignore_transform_lookup_error = true,
+		};
+
```
cagney: this feels like code duplicated from algparse?

Vinnu124: Yes , its similar to what algparse does.
I will modify it , so that it output does look better

Vinnu124:
@cagney , i have updated the patch to print it similar to the algparse format.
I shall update the tests if this seems fine.

Vinnu124:
Also have removed the debug option , so the proposals always get displayed upon startup.

cagney:
> I shall update the tests if this seems fine.

yes, like it (procedural nit, it's better to use text and backticks when showing output and not screen snapshots)

programs/pluto/plutomain.c
cagney: can the code, specifically the part, emitting the proposals, be shared?

Vinnu124: @cagney 
https://github.com/libreswan/libreswan/blob/main/programs/algparse/algparse.c#L121
this is where it prints in the alparse.
do you maybe want a shared helper for these two??

cagney: Ah, yes, I see why you did the copy.

The reason this is catching my eye is that there's `[EESP](https://datatracker.ietf.org/doc/draft-ietf-ipsecme-eesp/)`, a successor to `ESP` making its way through the IETF.  This means that tables like:
```
struct protocol {
	const char *name;
	struct proposal_parser *(*parser)(const struct proposal_policy *policy);
	bool pfs_vs_dh;
	bool (*alg_is_ok)(const struct ike_alg *alg, const struct logger *logger);
};

const struct protocol ike_protocol = {
	"ike", ike_proposal_parser, .pfs_vs_dh = false, .alg_is_ok = ike_alg_is_ike,
};

const struct protocol ah_protocol = {
	"ah", ah_proposal_parser, .pfs_vs_dh = true, .alg_is_ok = kernel_alg_is_ok,
};

const struct protocol esp_protocol = {
	"esp", esp_proposal_parser, .pfs_vs_dh = true, .alg_is_ok = kernel_alg_is_ok,
};

const struct protocol *protocols[] = {
	&ike_protocol,
	&ah_protocol,
	&esp_protocol,
};
```
will need to, at some point, be updated.  If we can avoid creating duplicates we're ahead.

Can I suggest first refactoring `protocols` as `struct ike_alg_protocol ike_alg_protocols[]` to ike_alg.[hc] (I'd terminate the array with NULL).  Then this code can use that table.

cagney (changes requested):

programs/pluto/plutomain.c
```diff
@@ -586,6 +587,56 @@ static deltatime_t check_config_deltatime(enum config_setup_keyword kw,
...
+			llog(RC_LOG, logger, "%sdefault IKEv%d %s proposals:",
+			     (is_fips_mode() ? "FIPS " : ""),
+			     protos[i].version,
+			     protos[i].name);
+			FOR_EACH_PROPOSAL(proposals, proposal) {
+				LLOG_JAMBUF(RC_LOG, logger, buf) {
+					jam_string(buf, "  ");
+					jam_proposal(buf, proposal);
+				}
+			}
```
cagney, line 634: Just an aside.  When it comes to this code displaying the output, the above looks good!
The things that matter - `proposals_from_str()` and `jam_proposal()` are both there - and the rest is boilerplate text that works no matter the number of protocols.

programs/pluto/plutomain.c
Vinnu124: Makes sense.
I'll  open a new PR for refactoring of `protocols`.(don't wanna cluster this up too much).

cagney: I'd just make the shuffling the first of two code commits (plus the test tweaks) in this PR.  Often changes come in two parts:
- re-arrange the deck chairs
something non-functional - has no affect on the testsuite
- make the actual change

lib/libswan/ike_alg.c
```diff
@@ -1425,3 +1425,48 @@ void init_ike_alg(struct logger *logger)
 	 */
 	log_ike_algs(logger);
 }
+/*
+ * Shared protocol table for IKE, ESP, and AH.
+ * Pluto can override alg_is_ok
+ * in the proposal_policy with the real kernel_alg_is_ok.
+ */
+
+static bool default_alg_is_ok(const struct ike_alg *alg,
```
Vinnu124, line 1434: This method would be overridden by pluto to actually check with the kernel (`kernel_alg_is_ok`)

Vinnu124:
@cagney
refactored the protocol parser from algparse.
Also have updated the display of proposals at startup in plutomain.
Is this what is intended? WDYT?

cagney (changes requested):
This is much better.  Is it possible to shuffle/merge the commits down to:

- refactor: move algparse protocol table into common ike_alg_protocols array
- pluto: log default proposals at startup

I'll kick off a test run.

Vinnu124:
Thanks!
Have updated the commits

cagney:
This is the only tweak I found
```
--- MAIN/testing/pluto/algparse-02-fips/west.console.txt
+++ OUTPUT/testing/pluto/algparse-02-fips/west.console.txt
@@ -2820,6 +2820,12 @@
 FIPS Key Exchange Method (DH):
 FIPS IP Compression:
 FIPS Sequence Number:
+FIPS default IKEv2 ike proposals:
+FIPS default IKEv1 ike proposals:
+FIPS default IKEv2 esp proposals:
+FIPS default IKEv1 esp proposals:
+FIPS default IKEv2 ah proposals:
+FIPS default IKEv1 ah proposals:
 west #
  # Check pluto algorithm list.
 west #
```

Vinnu124:
Thanks for the diff
Changed the test.
Rebased as well

PR #2811, opened 2026-05-27
feat(pluto): plant connection reqid in trap XFRMA_TMPL; capture from ACQUIRE

Vinnu124:
When installing a trap or shunt policy via add_spd_kernel_policy(), plant c->child.reqid into the XFRMA_TMPL attribute.
When the kernel fires the trap and sends ACQUIRE message , capture the reqid echoed back and store it in the policy_id.
This is groundwork to find the connection using hashtable.
Also update the existing tests where trap policies previously showed reqid 0.

### DIFF
--- a/programs/pluto/kernel_policy.c
+++ b/programs/pluto/kernel_policy.c
@@ -359,6 +359,14 @@ bool add_spd_kernel_policy(const struct spd *spd,
 					&nic_offload,
 					where);
 
+	/*
+	 * Plant the  reqid into the policy template so the kernel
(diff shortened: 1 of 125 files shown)


cagney (approved):

cagney:
Hopefully that is the painful testsuite change done.

Vinnu124:
Yess :)

cagney:
once the revival is tweaked I can merge

cagney (approved):
nice leap forward

PR #2826, opened 2026-06-03
Rename libexecdir to ipsec execdir

### LINKED ISSUE #2825: @@LIBEXECDIR@@ is not libexec!
### What needs to be done

It's actually libexec/pluto aka IPSEC_EXECDIR
(my bug)

cagney:
LIBEXECDIR points at something like /usr/libexec, but libreswan's build has it pointing at /usr/libexec/ipsec

close #2825


cagney:
it should be either IPSEC_LIBEXECDIR or PKGLIBEXECDIR; latter is what automake uses but former is more consistent with other make variables.

Issue #2097, opened 2025-03-16
cleanup ldap crl fetch support; was drop it?


cagney:
### What needs to be done

It isn't built in by default; and isn't tested.

cagney:
but it does seem to still build

bleve:
rhel and fedora build with LDAP, so it is at least built a lot.

We use lagacy interface for openldap, that should be upgraded.

bleve:
Checked fedora/rhel/debian, all those at least build curl with ldap support, so if we move fetching to separate program, we could at same time remove direct openldap use and rely on curl on that too.

cagney:
pkgsrc doesn't bundle ldap with curl by default (probably to avoid bloat); however should change this:
```
        if (startswith(url, "ldap:")) {
                ugh = fetch_ldap_url(url, blob, logger);
        } else {
                ugh = fetch_curl(url, blob, logger);
        }
```
to, when no USE_LDAP, always call fetch_curl() with the URL

PR #2827, opened 2026-06-03
documentation: rpm build requires openldap-devel

The-Mule:
Latest versions of Debian (13), Devuan (6) and Fedora (44) are shipped with curl supporting ldap/ldaps protocols and hence we no longer need to package them direct ldap support. This is not the case for RHEL since its ubi9 and ubi10 minimal versions come with libcurl-minimal lacking ldap protocol support.

_Original description was: RHEL and Fedora specfiles contain `USE_LDAP=true` and build requires `openldap-devel`._

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,14 @@ For Debian / Ubuntu / Mint
       libldns-dev xmlto libcurl4-openssl-dev \
       systemd-dev
 
-For Fedora/CentOS-Stream/RHEL/AlmaLinux/RockyLinux etc.
+
+For Fedora/CentOS-Stream/RHEL/AlmaLinux/RockyLinux etc.:
+
+    If you have source repositories enabled:
+
+    dnf builddep libreswan
+
+    Otherwise:
 
     dnf install audit-libs-devel bison curl-devel flex \
       gcc ldns-devel libcap-ng-devel libevent-devel \
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -26,7 +26,6 @@ override_dh_auto_build:
 		PREFIX=/usr \
 		LIBEXECDIR=/usr/libexec/ipsec \
 		MANDIR=/usr/share/man \
-		USE_LDAP=true \
 		INITSYSTEM=systemd \
 		$(ENABLE_LIBCAP_NG) \
 		$(ENABLE_SELINUX)
@@ -39,7 +38,6 @@ override_dh_auto_install-arch:
 		PREFIX=/usr \
 		LIBEXECDIR=/usr/libexec/ipsec \
 		MANDIR=/usr/share/man \
-		USE_LDAP=true \
 		INITSYSTEM=systemd \
 		$(ENABLE_LIBCAP_NG) \
 		$(ENABLE_SELINUX) \
--- a/packaging/devuan/rules
+++ b/packaging/devuan/rules
@@ -26,7 +26,6 @@ override_dh_auto_build:
 		PREFIX=/usr \
 		LIBEXECDIR=/usr/libexec/ipsec \
 		NSSDIR=/var/lib/ipsec/nss \
-		USE_LDAP=true \
 		$(ENABLE_LIBCAP_NG) \
 		$(ENABLE_SELINUX)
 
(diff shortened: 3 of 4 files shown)


bleve:
That is outdated info. Libreswan doesn't use openldap any more. It now uses libcurl.

The-Mule:
There's still:

https://github.com/libreswan/libreswan/blob/37add508d944b0510a0ead64d2fea2cbdfe7d22b/programs/_import_crl/import_crl.h#L34-L40

curl is used by default but there's still seems to be handle to use ldap...?

bleve:
See https://github.com/libreswan/libreswan/issues/2097 - when build without USE_LDAP, curl is used to pull from ldap. So better fix is to remove USE_LDAP and remove direct openldap dependency.

cagney:
The default is USE_LDAP=false.

Where things go off script is in the spec files:
```
./packaging/rhel/9/libreswan.spec:23:    USE_LDAP=true \\\
./packaging/rhel/8/libreswan.spec:23:    USE_LDAP=true \\\
./packaging/devuan/rules:29:		USE_LDAP=true \
./packaging/devuan/rules:41:		USE_LDAP=true \
./packaging/fedora/libreswan.spec:21:    USE_LDAP=true \\\
./packaging/debian/rules:29:		USE_LDAP=true \
./packaging/debian/rules:42:		USE_LDAP=true \
```

cagney:
> curl is used by default but there's still seems to be handle to use ldap...?

pkgsrc doesn't include ldap support in curl

The-Mule:
Also RHEL ubi image curl (libcurl-minimal) does not have ldap:

```
curl 8.12.1 (x86_64-koji-linux-gnu) libcurl/8.12.1 OpenSSL/3.5.5 zlib/1.3.1.zlib-ng libidn2/2.3.7 nghttp2/1.68.0
Release-Date: 2025-02-13
Protocols: file ftp ftps http https ipfs ipns
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz SPNEGO SSL threadsafe UnixSockets
```


On the other hand fedora-minimal image does not have the same problem:

```
curl 8.18.0 (x86_64-redhat-linux-gnu) libcurl/8.18.0 OpenSSL/3.5.5 zlib/1.3.1.zlib-ng brotli/1.2.0 libidn2/2.3.8 libpsl/0.21.5 libssh/0.12.0/openssl/zlib nghttp2/1.68.0 ngtcp2/1.22.1 nghttp3/1.16.0 mit-krb5/1.22.2 OpenLDAP/2.6.13
Release-Date: 2026-01-07
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets
```

Debian 13 / Devuan 6 curl supports ldap/ldaps as well so I guess the right solution is indeed just to keep USE_LDAP=true in RHEL specfiles and remove it from the other specs:

```
./packaging/debian/rules
./packaging/devuan/rules
./packaging/fedora/libreswan.spec
```

bleve:
Even on rhel there is no point in building with openldap, rhel curl is build with ldap support.

cagney:
So:

> Also RHEL ubi image curl (libcurl-minimal) does not have ldap:

and

> Even on rhel there is no point in building with openldap, rhel curl is build with ldap support.

is this comparing different RHELs?

> USE_LDAP=true in RHEL specfiles and remove it from the other specs:
```
./packaging/debian/rules
./packaging/devuan/rules
./packaging/fedora/libreswan.spec
```

yes.  For dependencies, I think the README should somehow steer the developer towards `dnf builddep`.

The-Mule:
> For dependencies, I think the README should somehow steer the developer towards dnf builddep.

Right, that's actually a pretty good idea (replacing `dnf install ...` for Fedora with `dnf builddep`).

README.md
```diff
@@ -47,7 +47,14 @@ For Debian / Ubuntu / Mint
...
 
-For Fedora/CentOS-Stream/RHEL/AlmaLinux/RockyLinux etc.
+
+For Fedora/CentOS-Stream/RHEL/AlmaLinux/RockyLinux etc.:
+
+    If you have source repositories enabled:
+
+    dnf builddep libreswan
+
+    Otherwise:
```
cagney, line 57: Sneaky, I like it!

cagney (approved):

PR #2839, opened 2026-06-07
xfrm larval PoC 

antonyantony:
This is PoC code. The PR is only for informational and to discuss the concept of larval skbs in kernel.

- did you use "AI" : YES

  We're assuming this pull-request is your, and your co-authors, work;
  and it made available under the terms of the GPLv2+.

- have you seen our
  [Programming Conventions](https://github.com/libreswan/libreswan/wiki/Hacking:-Programming-Conventions)
  guideline YES

- have you seen our
  [Git, GitHub, and Pull Requests](https://github.com/libreswan/libreswan/wiki/Hacking:-Git,-GitHub,-and-Pull-Requests)
  guideline YES


cagney:
@Vinnu124 and I were discussing this

I've a question on mechanics:

- pluto has trap for 192.0.0.0/24-192.1.2.0/24 installed

- program sends 192.0.0.0->192.1.2.3/udp/22

  + what will the "shunt" / acquire look like
     I'm guessing just 192.0.0.0->192.1.2.3/udp/22?

  + is packet 1:1 tied to the shunt?

  + what happens if there's a re-transmit

- another program sends 192.0.0.1->192.1.2.4/tcp/22

   + does this add a further shunt with a further packet?

The current code when transitioning from ondemand->negotiating installs a widened block.

antonyantony:
in kernel terms there is only policy, no shunts. the kernel  message has both  packet that triggered and policy that matched, and no assoicated SA. Here is ICMP example. UDP would look somthing similar.

  sel src 192.0.1.254/32 dst 192.0.2.254/32 proto icmp type 8 code 0 dev eth1
  policy src 192.0.1.0/24 dst 192.0.2.0/24 
   dir out priority 1757393 ptype main
        tmpl src 0.0.0.0 dst 0.0.0.0
                proto esp reqid 0 mode transport

My guess duplicate packets will be duplicates. Run test? 
When the goal is pass pluto should not add anything else, a.k.a leave it alone.

Issue #2849, opened 2026-06-11
use ;addke1=... syntax when logging proposals with added ke


cagney:
### What needs to be done

I suspect it may still sometimes use the old style.

PR #2851, opened 2026-06-16
Remove dash-positional addke and sn syntax, log as ;name=

### LINKED ISSUE #2838: drop ike=...-addke1-addke2-addke3 ... syntax
### What needs to be done

... which is currently hiding behind intermediate=yes
- not documented
- addke1=... is more explicit

ruiliio:
Fixes #2838 and #2849 

This PR removes the dash-positional addke and sn syntax: `ike=…-addke1-addke2…` and the unoffical `-sn-`, keeps explicit `;addke1=` and `;sn= `, and switches their logging to `;addke1=` and `;sn= `.  The changes:

  - drop the dash parse path for addke and sn, the `policy->addke` bit in proposals, and the `algparse --addke` flag.
  - `jam_proposal()` logs addke and sn as `;addke1=…` and `;sn= ` instead of the old dash style.
  - convert the addke test proposals to `;addke1=` and regenerate the console output.


cagney:
looking

include/proposals.h
```diff
@@ -96,7 +96,6 @@ struct proposal_parser {
 struct proposal_policy {
 	enum ike_version version;
 	bool pfs; /* For CHILD SA, use DH from IKE SA */
-	bool addke;	/* allow ADDKE* algorithms */
 	bool check_pfs_vs_ke;
```
cagney, line 99: spot on

cagney (changes requested):
The code still allows the very unofficial `-SN-` and that sits between KE and ADDKE1.  That should also go (sorry, missed that).

With that in mind, I'd change the comparisons to:
```
transform_type <= transform_type_ke
```
et.al., as that leaves no space and is clearer as to where the old style stop.s

ruiliio:
Folded the -SN- removal into the existing three commits, and the comparisons now use transform_type <= transform_type_ke in both the parser and jam.

cagney (approved):
looks good

PR #2871, opened 2026-06-28
feat: display reqid in connection status

Vinnu124:
Explicitly display the connection's reqid (shows the reqid when `ipsec status` is used).
Updated tests as well

### DIFF
--- a/programs/pluto/whack_connectionstatus.c
+++ b/programs/pluto/whack_connectionstatus.c
@@ -1046,6 +1046,15 @@ static void show_connection_status(struct show *s, const struct connection *c)
 		jam_connection_owners(buf, c, IKE_SA_OWNER_FLOOR, IKE_SA_OWNER_ROOF);
 		jam_connection_owners(buf, c, CHILD_SA_OWNER_FLOOR, CHILD_SA_OWNER_ROOF);
 	}
+	/*
+	 * display the reqid for a particular connection
+	 */
+	SHOW_JAMBUF(s, buf) {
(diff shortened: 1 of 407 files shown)


programs/pluto/whack_connectionstatus.c
```diff
@@ -1046,6 +1046,15 @@ static void show_connection_status(struct show *s, const struct connection *c)
 		jam_connection_owners(buf, c, IKE_SA_OWNER_FLOOR, IKE_SA_OWNER_ROOF);
 		jam_connection_owners(buf, c, CHILD_SA_OWNER_FLOOR, CHILD_SA_OWNER_ROOF);
 	}
+	/*
+	 * display the reqid for a particular connection
+	 */
+	SHOW_JAMBUF(s, buf) {
+		jam_string(buf, c->name);
+		jam_string(buf, ":  ");
+		jam(buf, " connection reqid: "PRI_REQID";",
```
cagney: No need for "connection ", it's redundant.
I see your formatting is consistent with what is immediately around, which is good.
I think a separate line is ok.

cagney (changes requested):
just a tweak that sed will fix

programs/pluto/whack_connectionstatus.c
Vinnu124: Updated the tests. thanks

cagney (approved):
two thumbs up

(but not to github, their web UI keeps crashing)

PR #2874, opened 2026-06-29
ikev2: support IKE SA rekeying with IKE_FOLLOWUP_KE exchange

### LINKED ISSUE #2831: support  IKE_FOLLOWUP_KE
### What needs to be done

rfc says it all, see #2766

ueno:
This adds support for IKE_FOLLOWUP_KE exchange, which optionally follows CREATE_CHILD_SA exchange to rekey SA. 
Currently it only supports IKE SA rekeying. It would be straightforward to support Child SA rekeying, but before going too far, I would like to hear any opinions on the approach.

The basic idea is:
- add 3 new larval states `REKEY_IKE_FOLLOWUP_KE_I0`, `REKEY_IKE_FOLLOWUP_KE_I1`, and `REKEY_IKE_FOLLOWUP_KE_R0`, transitioned from `REKEY_IKE_I1` or `REKEY_IKE_R0`
- only the first and the final IKE_FOLLOWUP_KE drives the state
- the intermediate secrets (`SK(1) ... SK(n)`) are represented as an array of `PK11SymKey *`; it's tempting to concatenate them while accumulating, but this is written so for a future PKCS#11 mechanism design

Fixes: #2831


include/ike_alg_prf_ikev2_ops.h
```diff
@@ -39,11 +39,13 @@ struct prf_ikev2_ops {
...
 				       struct logger *logger);
-	/* SKEYSEED = prf(SK_d (old), g^ir (new) | Ni | Nr) */
+	/* SKEYSEED = prf(SK_d (old), g^ir (new) | Ni | Nr [ | SK(1) ... | SK(n) ]) */
 	PK11SymKey *(*ike_sa_rekey_skeyseed)(const struct prf_desc *prf_desc,
 					     PK11SymKey *old_SK_d,
 					     PK11SymKey *new_ke_secret,
 					     const chunk_t Ni, const chunk_t Nr,
+					     size_t nr_additional_secrets,
+					     PK11SymKey **additional_secrets,
 					     struct logger *logger);
```
cagney, line 49: The comment should change `g^ir` to `SK(0)`, since g^ir is obsolete.

Should `new_ke_secret`, `nr_additional_secrets`, and `additional_secrets` be combined into a single const struct pointer?  That would better reflect the RFC.

I'd look at `table.h` as one possible approach.  Specifically, libreswan now has the __counted_by__() attribute sprinkled over the code.

cagney:
> the intermediate secrets (SK(1) ... SK(n)) are represented as an array of PK11SymKey *; it's tempting to concatenate them while accumulating, but this is written so for a future PKCS#11 mechanism design

Is this speculative or is there a draft spec?  Either way, the RFC has:

```
SKEYSEED = prf(SK_d, SK(0) | Ni | Nr | SK(1) | ... SK(n))
```

so the code is keeping with that design.  Per my other comment, what I would do is merge the g_ir and this new array into a single table so it matches this latest RFC's terminology.

cagney:
> I would like to hear any opinions on the approach.

I was expecting and I see the chaining transition code.

I was expecting more code sharing with ikev2_ike_intermediate.c, but having looked more carefully at that section of the RFC, I see that my expectation was flawed.  Instead separate offload code is needed.

Like IKE_INTERMEDIATE it uses the new ikev2_helper.[hc] - it kept the offload functions local to the exchange which hopefully made life easier.   Once this is merged I might revisit the interface - a crypto helper failing is always instant death so sft_status is probably overkill.

ueno:
@cagney thanks for the suggestions. I've added 3 sub-commits split from the main one:

1. ikev2: define emancipate_larval_ike_sa as global function
2. ikev2: generalize and export extract_ike_intermediate_v2KE
3. ikev2: log ADDKE transforms when IKE SA is established

Now the KE extraction code is shared between IKE_INTERMEDIATE an IKE_FOLLOWUP_KE.

programs/pluto/ikev2_parent.c
```diff
@@ -212,6 +212,11 @@ void llog_v2_ike_sa_established(struct ike_sa *ike, struct child_sa *larval)
 		jam_string(buf, " ke=");
 		jam_string(buf, ike->sa.st_oakley.ta_dh->common.fqn);
 
+		FOR_EACH_ITEM(ke, &ike->sa.st_oakley.ta_addke) {
+			jam_string(buf, " ke=");
```
cagney: Yes, this is definitely missing, good catch.
I think the message should be changed to ` addkeN=`?  That way the logs clearly show which KE is for which addkeN.
Since this will touch all the the existing ml-kem tests, and is orthogonal to IKE_FOLLOW_UP_KE, it deserves its own PR

cagney (changes requested):

include/ike_alg_prf_ikev2_ops.h
cagney: Did you have thoughts on this?

I think minimally `nr_additional_secrets` and `additional_secrets` should be combined.

ueno: Yes, using a dedicated structure compatible with `table.h` sounds like a good idea. I'm on it.

programs/pluto/ikev2_parent.c
ueno: Actually, the function `llog_v2_ike_sa_established` is only used by `emancipate_larval_ike_sa` after rekey. Given the existing ML-KEM tests don't do rekey, they do not need update.

I created #2880 anyway with the format change (`ke=` to `addkeN=`).

include/ike_alg_prf_ikev2_ops.h
ueno: This should be done now.

ueno:
Further improvement ideas:
- maybe `new_v2_child_sa` call could be invoked with `REKEY_IKE_FOLLOWUP_[IR][01]` iff an IKE_FOLLOWUP_KE exchange is needed and otherwise initialize it with `REKEY_IKE_[IR][01]` to simplify the state machine transitions?
- calling `symkey_addref` for all additional secrets might not be needed, if we just pass `struct prf_keys *` around

cagney:
Heads up

```
kvmrunner 2:27:36.05:   tests/good/errors/ASSERTION: 1: whack-rekey-ike-02-async-ikev2-addke
kvmrunner 2:27:36.05:   tests/good/errors/CORE: 1: whack-rekey-ike-02-async-ikev2-addke
```

I'll dig up the details

cagney:
Looks like during shutdown
```
Thread 1 (Thread 0x7f44ef417c00 (LWP 892)):
#0  0x000055d2f182ce40 in delete_state (st=0x7f44ebdaf028) at state.c:1000
#1  0x000055d2f17c770f in visit_connection_states (c=c@entry=0x7f44ebf2ab98, state_visitor=<optimized out>, state_context=state_context@entry=0x7ffc22949204, where=0x55d2f19cf648 <terminate_connection.here>) at visit_connection.c:882
#2  0x000055d2f1825b09 in terminate_all_connection_states (c=0x7f44ebf2ab98, where=0x7f44ebf2ab98) at terminate.c:260
#3  terminate_connection (c=0x7f44ebf2ab98, where=0x55d2f19caf58 <delete_every_connection.here>) at terminate.c:301
#4  0x000055d2f1825e50 in terminate_and_down_and_unroute_connections (c=c@entry=0x7f44ebf2ab98, where=where@entry=0x55d2f19caf58 <delete_every_connection.here>) at terminate.c:368
#5  0x000055d2f1826249 in terminate_and_delete_connections (c=0x7f44ebf2ab98, logger=logger@entry=0x55d2f19f8a00 <global_logger>, where=where@entry=0x55d2f19caf58 <delete_every_connection.here>) at terminate.c:486
#6  0x000055d2f17d07de in delete_every_connection (logger=0x55d2f19f8a00 <global_logger>) at whack_shutdown.c:158
#7  exit_epilogue (logger=logger@entry=0x55d2f19f8a00 <global_logger>) at whack_shutdown.c:188
#8  0x000055d2f17d0aec in server_stopped_callback (r=<optimized out>, logger=0x55d2f19f8a00 <global_logger>) at whack_shutdown.c:312
#9  0x000055d2f18385b7 in run_server (conffile=0x7f44eebf2fe8 "/etc/ipsec.conf", logger=0x55d2f19f8a00 <global_logger>) at server.c:1167
#10 0x000055d2f183200f in main (argc=4, argv=<optimized out>) at plutomain.c:1557
```
```
995/* intermediate */
996free_chunk_content(&st->st_v2_ike_intermediate.initiator);
997free_chunk_content(&st->st_v2_ike_intermediate.responder);
998
999/* followup ke */
1000TABLE_FOR_EACH(key, st->st_v2_ike_followup_ke.keys) {
1001symkey_delref(st->logger, "SK(n)", key);
1002}
1003pfreeany(st->st_v2_ike_followup_ke.keys);
1004
```

cagney:
For the test name, can I suggest changing it to something that matches the pattern:
```
testing/pluto/ikev2-intermediate-*
```
so that someone running:
```
./kvm check testing/pluto/*intermediate*
```
picks up this test.

The folk law (which I'm possibly making up as I type) is that whack-* tests exercise the `ipsec ...` comands users typically run; but here the test is something much deeper.

ueno:
> #0  0x000055d2f182ce40 in delete_state (st=0x7f44ebdaf028) at state.c:1000

Thank you for looking into it. I added a NULL check to fix it.
Will change the test name soon.

ueno:
Renamed the test to `ikev2-intermediate-10-rekey-ike`. I also added handling of failure response, reusing `process_v2_CREATE_CHILD_SA_failure_response`.

cagney:
> Thank you for looking into it. I added a NULL check to fix it. Will change the test name soon.

What was NULL?

Both ` TABLE_FOR_EACH()` and `symkey_delref()` should have handled that?

ueno:
> > Thank you for looking into it. I added a NULL check to fix it. Will change the test name soon.
> 
> What was NULL?
> 
> Both ` TABLE_FOR_EACH()` and `symkey_delref()` should have handled that?

The entire table (i.e., `ike->sa.st_v2_ike_followup_ke.keys`) can be NULL, as it's allocated on demand, but the previous code didn't check that and called `TABLE_FOR_EACH` on it.

cagney:
> > > Thank you for looking into it. I added a NULL check to fix it. Will change the test name soon.
> > 
> > 
> > What was NULL?
> > Both ` TABLE_FOR_EACH()` and `symkey_delref()` should have handled that?
> 
> The entire table (i.e., `ike->sa.st_v2_ike_followup_ke.keys`) can be NULL, as it's allocated on demand, but the previous code didn't check that and called `TABLE_FOR_EACH` on it.

Except, per above, TABLE_FOR_EACH() is checking for NULL:

```
#define TABLE_FOR_EACH(ITEM, TABLE)                                     \
        for (typeof((TABLE)->table[0]) *ITEM = ((TABLE) == NULL ? NULL : \
                                                &(TABLE)->table[0]);    \
             ITEM != NULL && ITEM < &(TABLE)->table[(TABLE)->len];      \
             ITEM++)
```

programs/pluto/state.c
```diff
@@ -996,6 +996,14 @@ void delete_state(struct state *st)
 	free_chunk_content(&st->st_v2_ike_intermediate.initiator);
 	free_chunk_content(&st->st_v2_ike_intermediate.responder);
 
+	/* followup ke */
+	if (st->st_v2_ike_followup_ke.keys != NULL) {
```
cagney: The `!= NULL` can be removed.  The problem is a use-after-free:
```
999		/* followup ke */
1000		if (st->st_v2_ike_followup_ke.keys != NULL) {
1001			TABLE_FOR_EACH(key, st->st_v2_ike_followup_ke.keys) {
1002				symkey_delref(st->logger, "SK(n)", key);
1003			}
1004			pfreeany(st->st_v2_ike_followup_ke.keys);
1005		}
1006	
1007		/* session resumption */
1008		pfreeany(st->st_v2_resume_session);
(gdb) print st->st_v2_ike_followup_ke.keys
```
```
FATAL: ASSERTION FAILED: pointer 0x7f4e65d1eff8 invalid, possible heap corruption or bad pointer (magic != LEAK_MAGIC and ~LEAK_MAGIC}) (pfree() +207 lib/libswan/alloc.c)
```

cagney (changes requested):

programs/pluto/state.c
ueno: I overlooked `(TABLE) == NULL ? NULL : ...` in the `TABLE_FOR_EACH` definition, sorry. I removed the NULL check, but haven't managed to reproduce the issue so far; is it intermittent or do I need a special setup to trigger it?

cagney:
For me, with the new test it seems to happen consistently during shutdown, at least under KVM.
btw, pmemory() explodes when passed what looks like a suspect pointer

ueno:
Yes, there was a missing clone; I've fixed it in the "fixup!" commit.

cagney (changes requested):
- test results are pretty much perfect, this just needs a tweak
```
diffs_command testing/pluto/whack-globalstatus-01
--- MAIN/testing/pluto/whack-globalstatus-01/west.console.txt
+++ OUTPUT/testing/pluto/whack-globalstatus-01/west.console.txt
@@ -71,6 +71,9 @@
 current.states.enumerate.REKEY_CHILD_I1=0
 current.states.enumerate.REKEY_IKE_R0=0
 current.states.enumerate.REKEY_CHILD_R0=0
+current.states.enumerate.REKEY_IKE_FOLLOWUP_KE_I0=0
+current.states.enumerate.REKEY_IKE_FOLLOWUP_KE_I1=0
+current.states.enumerate.REKEY_IKE_FOLLOWUP_KE_R0=0
 current.states.enumerate.ESTABLISHED_IKE_SA=0
 current.states.enumerate.ESTABLISHED_CHILD_SA=0
 current.states.enumerate.ZOMBIE=0
```
- I'd fold the `fix: ikev1: ...` into the original commit
  (unless you'd like that preserved for ever :)

ueno:
> * test results are pretty much perfect, this just needs a tweak
> 
> 
> ```
> diffs_command testing/pluto/whack-globalstatus-01
> --- MAIN/testing/pluto/whack-globalstatus-01/west.console.txt
> +++ OUTPUT/testing/pluto/whack-globalstatus-01/west.console.txt
> [...]
> ```

Thanks, applied.

>     * I'd fold the `fix: ikev1: ...` into the original commit
>       (unless you'd like that preserved for ever :)

Merged into the original commit.

Also rebased against the current main.

cagney (approved):
Looks good!  Final test run in progress.

I'm guessing next is extend testing (perhaps try to interop with strongswan).
Is any work needed for Child SAs?

cagney:
```
$ ./kvm recheck testing/pluto/*intermediate*
kvmrunner 5:33.07: Details:
kvmrunner 5:33.07: Summary:
kvmrunner 5:33.07:   tests/good/results/passed: 16
kvmrunner 5:33.07:   total: 16
kvmrunner 5:33.07:   total/passed: 16
kvmrunner 5:33.07: run finished at 2026-07-11 13:29:11.719923 after 0:05:33.751286
make: Leaving directory '/home/libreswan/git-misc/testing/kvm'
```

ueno:
I will follow-up on the Child SA rekey/creation. Given key exchanges for those are not mandatory, it is not a blocker for 5.4?

cagney:
Yes, I was thinking that too

PR #2895, opened 2026-07-07
pluto: reconcile connections on ipsec add --autoall

### LINKED ISSUE #2605: Support to reload full configuration without restarting Pluto
### What needs to be done

I would like to check if there are any plans to support reloading of the entire `ipsec.conf` configuration file without restarting Pluto. If not, I would like to kindly request for this feature to be added. We can have a new `ipsec` command to handle this case, maybe something like `ipsec reload` or `ipsec update`.

From [this comment](https://github.com/libreswan/libreswan/issues/2584#issuecomment-3785184089), I understand that we can specify `ipsec auto --start <conn_name>` to do this for a particular connection, but my query is more on doing this for the full configuration (which could potentially include multiple connections) without specifying the name of any connections.

This means that Libreswan should add new connections, remove deleted connections, restart existing connections with modified configuration, and keep active connections untouched if their configurations are unmodified.

From [this comment](https://github.com/libreswan/libreswan/issues/265#issuecomment-584304850), it seems that we currently do not keep track of the full configuration state on disk. I believe that if we want to support this feature, we would need to read the file when parsing the configuration, persistently store it in memory somewhere (maybe as a configuration struct), and when reloading is requested, we parse the current configuration file on disk and compare that to the stored configuration in memory to determine what has changed. We can use something like `inotify` if we want to automatically detect when the file has been changed, but that can be a stretch goal for now.

I am open to contribute some code and raise a PR to add this feature, but I would like to hear some thoughts from the community first. I would also require some pointers and assistance regarding the appropriate approach to implement this from the maintainers.

Thank you!

jamestiotio:
ipsec add --autoall (run when pluto starts or when ipsec.conf is reloaded) is meant to reconcile pluto's connection table against the current ipsec.conf, but it was additive only: connections removed from the config lingered in pluto as zombies, and unchanged connections were torn down and rebuilt on every reload, dropping their SAs.

Add a "config setup" keyword `autoall-reconcile`, which defaults to no, that opts in to reconciliation. When enabled, addconn and add bracket the per-conn WHACK_ADD stream with WHACK_AUTOALL_{START,STOP}:

  - START marks every root connection stale.
  - Each WHACK_ADD carries a digest of its config-derived fields, so pluto skips the delete and re-add when the digest matches (and clears the stale flag) and only deletes + re-adds when the config changed.
  - STOP sweeps any roots still flagged stale.

The subnets= permutations are matched via connalias so they participate in the optimization alongside their base_name root.

Adds tests covering the sweep, no-op, modified-conn, and disabled paths.

This closes #2605.

### DIFF
--- a/configs/d.ipsec.conf/setup/autoall-sweep.xml
+++ b/configs/d.ipsec.conf/setup/autoall-sweep.xml
@@ -1,0 +1,22 @@
+<varlistentry id='setup.autoall-sweep'>
+  <term>
+    <option>autoall-sweep=&yn_option;</option>
+  </term>
+  <listitem>
+    <para>
+      Whether <command>ipsec add --autoall</command> (run when &pluto;
+      starts and whenever <filename>ipsec.conf</filename> is reloaded)
+      should reconcile the running connection table against the
+      configuration rather than only adding to it.  When enabled,
+      connections whose definition is unchanged are left in place
+      (preserving their established SAs instead of being torn down and
+      rebuilt), and connections that no longer appear in the
+      configuration are swept away.  When disabled,
+      <option>--autoall</option> is additive only: removed connections
+      linger and unchanged connections are re-added on every reload.
+    </para>
+    <para>
+      The default is &no;.
+    </para>
+  </listitem>
+</varlistentry>
--- a/configs/ipsec.conf.5.xml
+++ b/configs/ipsec.conf.5.xml
@@ -145,6 +145,7 @@
 <!ENTITY conn.xauthserver SYSTEM "d.ipsec.conf/conn/xauthserver.xml">
 
 <!ENTITY setup.audit-log SYSTEM "d.ipsec.conf/setup/audit-log.xml">
+<!ENTITY setup.autoall-sweep SYSTEM "d.ipsec.conf/setup/autoall-sweep.xml">
 <!ENTITY setup.crl-strict SYSTEM "d.ipsec.conf/setup/crl-strict.xml">
 <!ENTITY setup.crl-timeout SYSTEM "d.ipsec.conf/setup/crl-timeout.xml">
 <!ENTITY setup.crlcheckinterval SYSTEM "d.ipsec.conf/setup/crlcheckinterval.xml">
--- a/lib/libswan/ipsecconf/starterwhack.c
+++ b/lib/libswan/ipsecconf/starterwhack.c
@@ -50,7 +50,8 @@ int starter_whack_add_conn(const char *ctlsocket,
 			   struct logger *logger,
 			   bool dry_run,
 			   enum yn_options async,
-			   enum whack_noise noise)
+			   enum whack_noise noise,
+			   bool check_message)
 {
 	struct whack_message msg;
 	init_whack_message(&msg, WHACK_FROM_ADDCONN);
@@ -58,6 +59,7 @@ int starter_whack_add_conn(const char *ctlsocket,
 	msg.whack_command = WHACK_ADD;
 	msg.name = conn->name;
 	msg.whack_async = (async == YN_YES);
+	msg.whack_check_message = check_message;
 
 	for (enum config_conn_keyword kw = 1; kw < CONFIG_CONN_KEYWORD_ROOF; kw++) {
 		msg.conn[END_ROOF].value[kw] = conn->values[kw].string;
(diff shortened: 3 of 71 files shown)


cagney:
thanks! it's in the queue; but it's northern summer so the queue is being a little slow

jamestiotio:
No worries, take your time!

lib/libswan/whack/pickle.c
```diff
@@ -481,6 +481,7 @@ err_t pack_whack_msg(struct whackpacker *wp, struct logger *logger)
 	if (!pickle_whack_message(wp, &pickle_packer, logger)) {
 		return "too many bytes of strings or key to fit in message to pluto";
 	}
+	wp->msg->str_size = wp->str_next - wp->msg->string;
```
cagney: I don't understand this change....

Ah, ok I've figured it out.  It's so the hash function knows how many bytes to hash.  But can't that be calculated from .str_roof?

configs/ipsec.conf.5.xml
```diff
@@ -145,6 +145,7 @@
 <!ENTITY conn.xauthserver SYSTEM "d.ipsec.conf/conn/xauthserver.xml">
 
 <!ENTITY setup.audit-log SYSTEM "d.ipsec.conf/setup/audit-log.xml">
+<!ENTITY setup.autoall-sweep SYSTEM "d.ipsec.conf/setup/autoall-sweep.xml">
```
cagney: Just a heads up (I guess for me).

This is going to clash when merged as mainline's been changed to generate these entries.  If you decide to rebase my life will be easier :-)

configs/d.ipsec.conf/setup/autoall-sweep.xml
```diff
@@ -0,0 +1,22 @@
+<varlistentry id='setup.autoall-sweep'>
+  <term>
+    <option>autoall-sweep=&yn_option;</option>
```
cagney: I was wondering about `--autoall=sweep`, however since there's the new config option `autoall-sweep=...` that makes no sense - best to keep the config name and the option name inline.

programs/add/add.c
```diff
@@ -434,6 +435,8 @@ int main(int argc, char *argv[])
 #endif
 
 	if (autoall) {
+		bool autoall_sweep = config_setup_yn(KYN_AUTOALL_SWEEP);
```
cagney: Neat.

lib/libswan/ipsecconf/starterwhack.c
```diff
@@ -50,14 +50,16 @@ int starter_whack_add_conn(const char *ctlsocket,
...
+			   enum whack_noise noise,
+			   bool check_message)
 {
 	struct whack_message msg;
 	init_whack_message(&msg, WHACK_FROM_ADDCONN);
 
 	msg.whack_command = WHACK_ADD;
 	msg.name = conn->name;
 	msg.whack_async = (async == YN_YES);
+	msg.whack_check_message = check_message;
```
cagney: I think this needs a different name - something to indicate that a sweep is going on (it isn't verifying the whack message).

programs/pluto/whack_add.c
```diff
@@ -367,6 +440,28 @@ void whack_add(struct whack_message_refcnt *wmr, struct show *s)
 		 * This is old-to-new which means that aliases are processed
 		 * before templates.
 		 */
+		if (wm->whack_check_message) {
```
cagney: Can you please add comments here explaining whats happening.

cagney (changes requested):
First the cosmetic thing, there should be three commits:

- code
- documentation
- testsuite

I like the subject/description - I can cut/paste them into the merge message.

The rest is just cosmetic - once I dug through the code there were no real surprises.

lib/libswan/whack/pickle.c
jamestiotio: I have updated the code to use `str_roof` instead.

configs/ipsec.conf.5.xml
jamestiotio: I have rebased my changes on top of latest commit of the main branch. I believe this is taken care of now.

programs/pluto/whack_add.c
jamestiotio: I have updated the comments for this code block.

configs/d.ipsec.conf/setup/autoall-sweep.xml
jamestiotio: I have renamed this option to `autoall-reconcile` since technically, it enables both the marking and the sweeping, not just the sweeping.

lib/libswan/ipsecconf/starterwhack.c
jamestiotio: I have renamed this.

jamestiotio:
I have split the changes into the 3 commits as requested and incorporated your feedback @cagney. I have also refactored the code a bit so that it attempts to resolve to the actual IP for the cases of DNS, default route, and interface names. If they resolve to the same actual IP, then the connection is unchanged. Otherwise, the connection will be torn down and then rebuilt.

cagney:
> I have also refactored the code a bit so that it attempts to resolve to the actual IP for the cases of DNS, default route, and interface names. If they resolve to the same actual IP, then the connection is unchanged. Otherwise, the connection will be torn down and then rebuilt.

This needs a bit of brainstorming.

The current approach is to not break what isn't broken.  Only resolve DNS while trying to establish, or after the connection fails.  The default route should be similar.

Some of this is new to 5.4.  The closed bugs in https://github.com/orgs/libreswan/projects/11 provide some background.  But one specific problem is that, in some configurations, DNS can only resolve when the connection is down.

PR #2902, opened 2026-07-10
testing:sanitize auto generated reqids only

### LINKED ISSUE #2891: only sanitize auto-generated REQIDs in test output
### What needs to be done

for instance
```
... reqid: 100
```
which would come from the config `reqid=100` shouldn't be sanitized, but anything greater than ??? should be

this is similar to port numbers, where only ephemeral ports (>32k) get sanitized

by doing this we can see when a reqid is configured vs generated/non deterministic

Vinnu124:
Resolves #2891

Auto Generated reqid starts after  0x3fff(16383).
Sanitizing all reqids with length greater than or equal to 5

### DIFF
--- a/testing/pluto/flush-on-startup-xfrm/west.console.txt
+++ b/testing/pluto/flush-on-startup-xfrm/west.console.txt
@@ -21,12 +21,12 @@ west #
 west #
  ipsec _kernel state
 src 2.2.2.2 dst 1.1.1.1
-	proto esp spi 0xSPISPI reqid REQID mode tunnel
+	proto esp spi 0xSPISPI reqid 1000 mode tunnel
 	replay-window 0 
 	aead rfc4106(gcm(aes)) 0xENCAUTHKEY 128
 	sel src 0.0.0.0/0 dst 0.0.0.0/0 
 src 1.1.1.1 dst 2.2.2.2
-	proto esp spi 0xSPISPI reqid REQID mode tunnel
+	proto esp spi 0xSPISPI reqid 1000 mode tunnel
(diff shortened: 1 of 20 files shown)


testing/pluto/flush-on-startup-xfrm/west.console.txt
```diff
@@ -21,12 +21,12 @@ west #
 west #
  ipsec _kernel state
 src 2.2.2.2 dst 1.1.1.1
-	proto esp spi 0xSPISPI reqid REQID mode tunnel
+	proto esp spi 0xSPISPI reqid 1000 mode tunnel
```
Vinnu124, line 24: Has a hardcoded Id as 1000(deterministic)

testing/pluto/interop-ikev1-strongswan-10-ah-initiator-sha256/west.console.txt
```diff
@@ -66,13 +66,13 @@ grep: strongswan: No such file or directory
 west #
  ipsec _kernel state
 src 192.1.2.45 dst 192.1.2.23
-	proto ah spi 0xSPISPI reqid REQID mode tunnel
+	proto ah spi 0xSPISPI reqid 1 mode tunnel
```
Vinnu124, line 69: strongswan tests at default start their counter from 1(still deterministic for single connections)

cagney: ok

cagney:
testing

PR #2903, opened 2026-07-10
pfkeyv2: replace outbound kernel policy when one is already installed

aamrindersingh:
see the pfkeyv2-ondemand tests (wip #netbsd-wip)

On NetBSD the outbound policy install fails with EEXIST because the ondemand policy is still there, and the connection gets torn down, use REPLACE instead of ADD when an outbound policy is already installed

Also updates the transport test to use host to host selectors so transport mode can actually work. Both tests now pass

### DIFF
--- a/testing/pluto/pfkeyv2-ondemand-transport/all.netbsdwest-linuxeast.sh
+++ b/testing/pluto/pfkeyv2-ondemand-transport/all.netbsdwest-linuxeast.sh
@@ -16,6 +16,6 @@ west# ipsec _kernel state
 west# ipsec _kernel policy
 
 # trigger acquire
-west# ../../guestbin/ping-once.sh --fire-and-forget -I 192.0.1.254 192.0.2.254
-west# sleep 5 # negotiate
-west# ../../guestbin/ping-once.sh --up -I 192.0.1.254 192.0.2.254
+west# ../../guestbin/ping-once.sh --fire-and-forget 192.1.2.23
+west# ../../guestbin/wait-for-pluto.sh '^".*#2: initiator established Child SA'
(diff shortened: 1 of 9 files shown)


cagney (changes requested):
wow, just some reorg tweaks

- the routing.[hc] additions can be a separate patch; neat idea
- use pfkeyv2: and not kernel: for the pfkey changes subject; so its clear it isn't generic code being touched

testing/pluto/pfkeyv2-ondemand-transport/all.netbsdwest-linuxeast.sh
```diff
@@ -16,6 +16,6 @@ west# ipsec _kernel state
 west# ipsec _kernel policy
 
 # trigger acquire
-west# ../../guestbin/ping-once.sh --fire-and-forget -I 192.0.1.254 192.0.2.254
+west# ../../guestbin/ping-once.sh --fire-and-forget 192.1.2.23
 west# sleep 5 # negotiate
-west# ../../guestbin/ping-once.sh --up -I 192.0.1.254 192.0.2.254
```
cagney, line 21: the
```
sleep 5
```
should be replaced with:
```
../../guestbin/wait-for-pluto.sh '^".*#2: IPsec SA established'
```
on old slow h/w this can take >5s

cagney (changes requested):

testing/pluto/pfkeyv2-ondemand-transport/all.netbsdwest-linuxeast.sh
aamrindersingh: Done, for ikev2 logs its "initiator established Child SA"

PR #2910, opened 2026-07-15
testing: verify if the hashtable route correctly selects connection

### LINKED ISSUE #574: acquire can pick the wrong connection
consider
   conn distraction
       ...
   conn west-to-east
      also=distraction
      rightsubnet=0.0.0.0/0
where both are added but only west-to-east is loaded.

acquire picks the first (technically last, new2old) connection, and not the one that installed the acquire

Vinnu124:
Earlier, the connection which does not install the kernel policy(`conn distraction`) would be selected if added after the actual connection which loads it(`conn west-to-east`).
The hashtable route now selects the connection strictly on the basis of reqid(provided by ACQUIRE) and estabilishes the correct connection.
Resolves #574


cagney (changes requested):
can you just tweak the description to name the bug this demonstrates as fixed

Vinnu124:
Updated. thanks

PR #2928, opened 2026-07-20
testing: rename, refactor and simplify pluto/ikev2-x509-ecdsa-04-also-rsa/ test

vukasink:
in testing/pluto/ikev2-x509-ecdsa-04-also-rsa. Currently, test times out.


cagney:
ah, so the test expects the RSA key to come in via --x509 with the ECDSA key to be loaded explicitly

vukasink:
I am not sure the test expects RSA key at all, at least I am not reading it like that. I am understanding it like rsa is given as an option (in addition to ecdsa) in `authby`, but only ECDSA keys are loaded. The connection should be established, but it is not.

cagney:
in that case --nokeys is the correct option - init the NSS database but don't load keys

vukasink:
thanks, I'll update it, and at the same time cleanup this test (and `05-rsa-asym`, `05-rsa-asym-flip`)

cagney:
> which means it matches the ECDSA certificate (the one from hostname.p12 with libreswan test framework DNs) by it's nickname. I am not sure why the east/west.conf have strongswan DNs for {left,rightid} though, that's maybe the first point for cleanup.

probably, at the time the test was added - '22 - that's all there was

cagney:
btw leftid=%fromcert may be useful; but it doesn't explain why leftid=... was seemingly ignored
perhaps having an ID different to the cert's DN is a feature

vukasink:
okay, renamed and simplified the pluto/ikev2-x509-ecdsa-04-also-rsa/ test. Please feel free to rename it.

Notes:
1. I will make an issue regarding this test
2. I will update ikev2-x509-ecdsa-05-rsa-asym and ikev2-x509-ecdsa-06-rsa-asym-flip in a separate PR. These two tests (as I highly suspect ATM) are not buggy, they are just not properly setup. They are about asymmetric authentication where both sides have correct configs. But it's a separate issue/PR from this one.

vukasink:
> btw leftid=%fromcert may be useful; but it doesn't explain why leftid=... was seemingly ignored perhaps having an ID different to the cert's DN is a feature

yeah, I didn't look into that, it's an investigation for the future :)

vukasink:
cross-referencing the bug pertaining to this test: #2931

vukasink:
updated TESTLIST now

vukasink:
Will open new PR where both these changes and fixes to `ikev2-x509-ecdsa-05-...` and `ikev2-x509-ecdsa-06-...` are contained. I made an error in git branching steps.

cagney:
yea, just found:
```
[cagney@bernard git-hub main↑2]$ ./kvm failed
results_command --skip passed --skip unsupported --skip untested /home/libreswan/git-hub/testing
kvmresults 0.02: ****** /home/libreswan/git-hub/testing/pluto/TESTLIST:670: invalid test ikev2-x509-ecdsa-04-also-rsa: test directory not found: /home/libreswan/git-hub/testing/pluto/ikev2-x509-ecdsa-04-also-rsa
Traceback (most recent call last):
```
btw, what branching misstep?

vukasink:
I wanted to make the two blocks of changes on separate branches. I however branched the fixes to not-buggy `x509-ecdsa-05` and `x509-ecdsa-06` tests from this branch (and not from main). And in the meantime discovered I didn't update the TESTLIST. easiest solution then was to do a short rebase and submit new PR with everything in :)

PR #2932, opened 2026-07-20
testing: generate the host-to-host tests

### LINKED ISSUE #2868: migrate pfkeyv2 tests to east/west?
### What needs to be done

most simple tests use east-west but for pfkeyv2, rise-set are used; this came about because of historic limitations with the kvm framework

to make tests more recognisable we might as well update these tests.

aamrindersingh:
--  Adds testing/utils/host-to-host.sh (like check-10-topology.sh, only linux and the BSDs) to generate the host to host   transport , tunnel tests

-- Moves the existing tests from rise/set to east and west so rise/set are free to sit behind the gateways , upcoming :-)  

--Adds the missing host-to-host-{transport,  tunnel}-linux tests


cagney (approved):
two thumbs up

PR #2939, opened 2026-07-22
testing: add the host-to-host-tunnel-forward tests

aamrindersingh:
-- Adds host-to-host-tunnel-forward-{freebsd,linux,netbsd,openbsd} , rise pings set through the east=west tunnel (rise-east=TUNNEL=west-set) and back

-- on linux the forwarded packet goes through the fwd policy , the BSDs don't install one and forwarding still works (goldens show the difference)

-- generated by testing/utils/host-to-host.sh (easy)


cagney (approved):
As you can see on IRC, I've been trying stuff out.
Nice!

Issue #2896, opened 2026-07-09
expand ipsec.conf.5 to be clearer regarding non-RSA certificate usage


vukasink:
### What needs to be done

{left,right}cert= option in ipsec.conf man pages states `If you are using leftrsasigkey=%cert this defines the certificate nickname of your certificate in the NSS database. This can be on software or hardware security device.`, making me believe the option is only to be used with RSA-based certificates. However, it is used in ECDSA signatures (e.g., see test ikev2-x509-ecdsa-01), and should also be usable for EDDSA signatures.

@cagney states there are undocumented left/right ecdsakey, eddsakey and pubkey conf options, where "the preferred one should probably be {left,right}pubkey=".

Ultimately, the task here is: document ecdsakey, eddsakey and pubkey conf options, update {left,right}cert= option description and potentially make a test for {left,right}pubkey ?

cagney:
some background

As well as %cert, et.al. {left,right}{ecdsa,rsa,eddsa}sigkey= accept raw public keys.  The format is defined by the RFCs:
```
        IPSECKEY_ALGORITHM_RSA = 2,             /* RFC 4025 */
        IPSECKEY_ALGORITHM_ECDSA = 3,           /* RFC 8005 (see IANA Considerations) */
        IPSECKEY_ALGORITHM_EDDSA = 4,           /* RFC 9373 */
```
where as {left,right}pubkey= accepts a der encoded public key info
```
        IPSECKEY_ALGORITHM_X_PUBKEY = 5,        /* Subject Public Key Info - guess */
```
("guess" really means determine by examining the encoded ASN.1).

btw it is tested, see testing/pluto/whack-hostkey-07-ikev2-ckaid-pubkey-ecdsa which, imnsho is pretty evil

cagney:
hmm, the config file in the above test should contain a breadcrumb like:
```
#leftpubkey=<filled-in-later>
```
so it is found by grep

Issue #2919, opened 2026-07-18
generate or cross check ipsecconf/conn.c


paulwouters:
### Please Tell Us About Your System Configuration

what is pfs-rekey-workaround= 

### Please Tell Us What Happened?

It's not documented what this is supposed to do

paulwouters:
apparently:

```
 * When PFS_REKEY_WORKAROUND, the original ESP= line, instead of the
 * accepted_proposal, is used.
```

cagney:
a shallow check turns up
```
$ awk -F '"' '/K.*"/ { print $2 }' lib/libswan/ipsecconf/conn.c | while read o ; do test -r configs/d.ipsec.conf/conn/$o.xml || echo $o ; done

ecdsakey
eddsakey
pubkey
groundhog
also
connalias
ikev2
session-resumption
send-esp-tfc-padding-not-supported
iptfs-fragmentation
iptfs-packet-size
iptfs-max-queue-size
iptfs-reorder-window
iptfs-init-delay
iptfs-drop-time
pfs-rekey-workaround
dns-match-id,
ipsec-lifetime
sec-label
modecfgdns
modecfgdomains
modecfgbanner
clones
reauth
phase2alg
ikev1-dpdtimeout
```
- per #2896, the *key* ones are known
- some are probably hiding in other files and need to be teased out
- some, such as clones, are wip
- some, such as groundhog we may not want to admit

cagney:
this is a more up-to-date list
```
parameter ecdsakey is not documented
parameter eddsakey is not documented
parameter pubkey is not documented
parameter groundhog is not documented
parameter connalias is not documented
parameter session-resumption is not documented
parameter iptfs-packet-size is not documented
parameter pfs-rekey-workaround is not documented
parameter dns-match-id, is not documented
parameter clones is not documented
parameter reauth is not documented
```

PR #2945, opened 2026-07-24
patches for {pub,rsasig,ecdsa,eddsa}key documentation

vukasink:
Two commits:

- one fixing ipsec.conf (Issue #2896). I took artistic liberty here, and put everything under one option which is listed as `left{pub,rsasig,ecdsa,eddsa}key` (saves a lot of space)
- one updating some details in man pages of `ipsec {ecdsa,eddsa}sigkey` programs.

### DIFF
--- a/configs/d.ipsec.conf/conn/pubkey.xml
+++ b/configs/d.ipsec.conf/conn/pubkey.xml
@@ -1,0 +1,76 @@
+<varlistentry id='conn.pubkey'>
+  <term>
+    <option>leftpubkey=</option>
+  </term>
+  <term>
+    <option>rightpubkey=</option>
+  </term>
+  <listitem>
+    <para>
+      The left participant's public key for signature (either RSA, ECDSA or EdDSA)
+      authentication. The <option>leftpubkey</option> is the generic option.
+      The key should be in base-64 encoded RFC 2537 format (with 0s prepended),
+      or one of the following:
+      <variablelist>
+	<varlistentry>
+	  <term>
+	    <option>%none</option>
+	  </term>
+	  <listitem>
+	    <para>
+	      the same as not specifying a value (useful to override a
+	      default)
+	    </para>
+	  </listitem>
+	</varlistentry>
+	<varlistentry>
+	  <term>
+	    <option>%dnsondemand</option>
+	  </term>
+	  <listitem>
+	    <para>
+	      (the default) the key is to be fetched from DNS at the
+	      time it is needed.
+	    </para>
+	  </listitem>
+	</varlistentry>
+	<varlistentry>
+	  <term>
+	    <option>%dns</option>
+	  </term>
+	  <listitem>
+	    <para>
+	      Treated as <option>%dnsondemand</option>.  The identity
+	      used for the left participant must be a specific host,
+	      not <option>%any</option> or another magic value.
+	    </para>
+	  </listitem>
+	</varlistentry>
+	<varlistentry>
+	  <term>
+	    <option>%cert</option>
+	  </term>
+	  <listitem>
+	    <para>
+	      the information required from a certificate defined in
+	      <option>%leftcert</option> and automatically define
+	      leftid for you
+	    </para>
+	  </listitem>
+	</varlistentry>
+      </variablelist>
+    </para>
+    <para>
+      Additionally, one can provide: a DER-encoded public key, or a raw
+      public key generated using <command>ipsec {rsasig,ecdsa,eddsa}key</command>
+      command.
+    </para>
+    <caution>
+      <para>
+	If two connection descriptions specify different public keys
+	for the same <option>leftid</option>, confusion and
+	madness will ensue.
+      </para>
+    </caution>
+  </listitem>
+</varlistentry>
--- a/configs/d.ipsec.conf/conn/cert.xml
+++ b/configs/d.ipsec.conf/conn/cert.xml
@@ -7,9 +7,10 @@
   </term>
   <listitem>
     <para>
-      If you are using <option>leftrsasigkey=%cert</option>
-      this defines the certificate nickname of your certificate in the NSS
-      database. This can be on software or hardware security device.
+      If you are using <option>left{pub,rsasig,ecdsa,eddsa}key=%cert</option>/
+      <option>right{pub,rsasig,ecdsa,eddsa}key=%cert</option>, this defines
+      the certificate nickname of your certificate in the NSS database.
+      This can be on software or hardware security device.
     </para>
   </listitem>
 </varlistentry>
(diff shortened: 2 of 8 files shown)


vukasink:
the approach for changes in ipsec.conf probably hinders simple solution for #2919

cagney:
yep, mainline is printing
```
parameter rsasigkey is not documented
parameter ecdsakey is not documented
parameter eddsakey is not documented
parameter groundhog is not documented
parameter connalias is not documented
parameter session-resumption is not documented
parameter iptfs-packet-size is not documented
parameter pfs-rekey-workaround is not documented
parameter dns-match-id, is not documented
parameter clones is not documented
parameter reauth is not documented
```

I don't think the contraction `left{pub,rsasig,ecdsa,eddsa}key=` helps.

- so that all connection parameters are searchable, they need to appear in ipsec.conf(5) using all possible canonical forms

  with the above `rightecdsakey=` won't be found

- each key has its own format
  ```
  enum ipseckey_algorithm_type {
        IPSECKEY_ALGORITHM_DSA = 1,             /* RFC 4025 */
        IPSECKEY_ALGORITHM_RSA = 2,             /* RFC 4025 */
        IPSECKEY_ALGORITHM_ECDSA = 3,           /* RFC 8005 (see IANA Considerations) */
        IPSECKEY_ALGORITHM_EDDSA = 4,           /* RFC 9373 */
        IPSECKEY_ALGORITHM_X_PUBKEY = 5,        /* Subject Public Key Info - guess */
  };
  ```
  so the'll need to be an entry for each key type clarifying this

What about creating a new sub-section similar to [CONN PARAMETERS: AUTOMATIC KEYING](https://testing.libreswan.org/latest/documentation/ipsec.conf.5.html#conn_parameters__automatic_keying)?
It could open with generic comments and discussion of things like %... then finally list the individual key types and their specifics (which should be really short).

BTW, since both leftrsasigkey and rightrsasigkey are mentioned the use of just left.. seems to struggle (but I have no good solution, local and peer sometimes help).

vukasink:
Thanks, that's a good idea. Should this new sub-section go above the `CONN PARAMETERS: DNS` subsection? Or somewhere else?

Another option could be, no new subsection, list all the generic arguments in left/rightpubkey, and rsasigkey,ecdsakey,eddsakey could point to ~"accept generic arguments listed in leftpubkey and rightpubkey, and additionally *stuff specific for this algorithm, with details of it*"

Regarding using just left..., yeah, couple of options use this approach (leftid, leftrsasigkey used it, leftcert). I'll change that as well

cagney:
> Thanks, that's a good idea. Should this new sub-section go above the `CONN PARAMETERS: DNS` subsection? Or somewhere else?

Anywhere is better, "automatic keying" is clearly a dumping ground; immediately after?

> Another option could be, no new subsection, list all the generic arguments in left/rightpubkey, and rsasigkey,ecdsakey,eddsakey could point to ~"accept generic arguments listed in leftpubkey and rightpubkey, and additionally _stuff specific for this algorithm, with details of it_"

True.

On that, the ID gets used to generate an anchor in the html (we just can't see it).  There's probably a DocBook way of generating a reference to the anchor.

> Regarding using just left..., yeah, couple of options use this approach (leftid, leftrsasigkey used it, leftcert). I'll change that as well

Yes.  It goes back to when only the left parameter was documented.   I suspect it carried the assumption that left was local.

vukasink:
> True.
> 
> On that, the ID gets used to generate an anchor in the html (we just can't see it). There's probably a DocBook way of generating a reference to the anchor.
> 
This option with no new subsection seems cleaner to me. But I don't understand this part, since I am not familiar with it. I think I get your point regarding man pages that are in the html format on libreswan.org domain (e.g. [https://testing.libreswan.org/latest/documentation/ipsec.conf.5.html](IPSEC.CONF)), but even if what you're suggesting is possible, clicking a link in terminal running `man ipsec.conf` would still not be possible? Disclaimer: I have no idea, I am trying to learn and undestand what's possible and what's not.

cagney:
> This option with no new subsection seems cleaner to me.

m'kay

> But I don't understand this part, since I am not familiar with it. I think I get your point regarding man pages that are in the html format on libreswan.org domain (e.g. [https://testing.libreswan.org/latest/documentation/ipsec.conf.5.html](IPSEC.CONF)), but even if what you're suggesting is possible, clicking a link in terminal running `man ipsec.conf` would still not be possible? Disclaimer: I have no idea, I am trying to learn and undestand what's possible and what's not.

That's correct.  It would only affect output such as HTML or PDF.  Leave that DocBook battle for another day.

vukasink:
> > This option with no new subsection seems cleaner to me.
> 
> m'kay
> 

this is maybe a cultural and generational thing, but I read that as not agreeing fully :smiley: to be clear, I am fine with both options, let me know your preference between the new subsection and no new subsection, and I'll update the commit :)

cagney:
search for "M'kay South Park"; no implied meaning

cagney:
FYI, I added links.

vukasink:
I gave it another try, I'm open for more suggestions. Did not add references now because of #2948

cagney:
> Did not add references now because of #2948

I'm not sure I follow.

vukasink:
Ah, sorry :smiley:  I meant, "Did not add references now because it seems resolving #2948 will set an example of how to reference in different (referencing) cases, so it seems better to delay adding references here now". I can add references, though. Especially if there'll be another iteration over the PR :)

configs/d.ipsec.conf/conn/pubkey.xml
```diff
@@ -0,0 +1,91 @@
+<varlistentry id='conn.pubkey'>
+  <term>
+    <option>left{pub,rsasig,ecdsa,eddsa}key=</option>
+  </term>
```
cagney, line 4: Does this need tweaking?

configs/d.ipsec.conf/conn/pubkey.xml
```diff
@@ -0,0 +1,91 @@
...
+	    <option>%dns</option>
+	  </term>
+	  <listitem>
+	    <para>
+	      currently treated as <option>%dnsonload</option> but
+	      will change to <option>%dnsondemand</option> in the
+	      future.  The identity used for the left participant must
+	      be a specific host, not <option>%any</option> or
+	      another magic value.
+	    </para>
```
cagney, line 46: Looks like the switch happened.  I can't find dnsonload in the code base.

In fact, the only reference I found with a quick grep was:
```
v2.0/programs/auto/auto.in:		if (rk == "%dns" || rk == "%dnsonload")
```
my best guess is  1afc267b19e4a2f97a066aec15410bbbc061290f or 2.5, but that's so long ago as to be meaningless.

I'd just drop the reference to %dnsonload.

cagney (changes requested):
Just two tweaks.

vukasink:
Fixed, thanks for the sharp eye!

cagney (approved):
slick! I can find documentation on pubkey=!

PR #2950, opened 2026-07-28
Add support for announcing supported authentication methods (RFC 9593)

OsemaFadhel:
Implement RFC 9593 SUPPORTED_AUTH_METHODS notification to announce supported authentication methods.

- add emit_v2N_SUPPORTED_AUTH_METHODS() and parse_v2N_SUPPORTED_AUTH_METHODS() to emit and process the notification
- update ikev2_ike_intermediate to emit and process the notification in the KE_INTERMEDIATE exchange
- update local_v2_auth() to select an auth method based on received SUPPORTED_AUTH_METHODS notification from peer

Testing:

- add SUPPORTED_AUTH_METHODS KVM tests covering the new option and notification logic


OsemaFadhel:
The tests are marked as wip because they pre-select the right certs using left/rightcert= to successfully authenticate, which bypasses the logic of dynamically choosing one, because setting leftcert means we already know which method is going to be selected.

cagney:
@OsemaFadhel I've run the testsuite, things seem ok.

I'd like to merge in several parts (probably two), can you please break out these (effectively) non-functional commits:

```
ikev2: add send-supported-auth-methods option
documentation: document send-supported-auth-methods in ipsec.conf
ikev2: register SUPPORTED_AUTH_METHODS notification
```

and then add a test similar to the other `addconn-??-*` tests to demonstrate that the value appears in the `ipsec connectionstatus` output.  Some things to note:
- older tests use lots of `conn` sections, but new tests tend to use `ipsec addconn --name ... supported-auth-methods=yes`.
- this will affect the output of other tests so there'll need to be other updates

I see the option is added to whack - I'm truly nonplussed as while it is still expected I see more value in encouraging the still experimental `ipsec addconn ...` command.  Anyway, since it is there it should be mentioned in ipsec-whack.8.

One question.  Should the new config field live in `struct config` (where we find fields added in the distant past) or `struct host_config` (which only has fields added recently) since it is "host" aka IKE aka AUTH thing.  If the IKE notification negotiation code is going to be re-worked it's a bit of a toss up; on the other hand, "always going forward, never in reverse".  Thoughts?

Anyway, with that it will should just merge.

This way the output churn from adding the parameter doesn't become a distraction to the real commits.

configs/d.ipsec.conf/sect/conn.xml
vukasink, line 1: I guess it's better to put the option below `authby` or `{left,right}auth`

programs/pluto/ikev2_supported_auth.c
```diff
@@ -0,0 +1,327 @@
...
+	 *						 1                   2
+	 *	 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3
+	 *	+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+	 *	|  Length (=3)  |  Auth Method  |   Cert Link   |
+	 *	+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+	 *
+	 */
+
+	if (authby.rsasig_v1_5) {
+		uint8_t ann3[3] = { THREE_OCTET_ANNOUNCEMENT_LENGTH,
```
vukasink: I missed this. Maybe the `[3]` should also be `[THREE_OCTET_ANNOUNCEMENT_LENGTH]` and not a magic value? (and similarly for other cases). 

Maybe cagney can chip in with an opinion.

cagney: True.  It's also more robust than [] which relies on the constructor having the correct number of fields.

testing/x509/generate.sh
```diff
@@ -566,4 +566,9 @@ real/mainca  west_chain_int_2           west_chain_endcert         1 1 1 /  digi
 # Cert chain with intermediate BC CA=Y missing
 real/mainca  mainca                     west-bc-missing-chain-int  1 1 1 /  certSigning,critical  /
 real/mainca  west-bc-missing-chain-int  west-bc-missing-chain-end  1 1 1 /  /                     /
+# Cert for SUPPORTED_AUTH_METHODS testing
```
cagney: Is this to get around the problem where having both an ec and an rsa cert with the same nicname in the NSS DB isn't possible?

If that's the case, does certutil's --rename help?  It probably means hacking import.sh instead.

Anyway, and either way, if this appears as a separate pull request it will go straight in.

testing/pluto/TESTLIST
```diff
@@ -1847,3 +1847,13 @@ kvmplutotest	ikev2-routes-subnets-singular		good	github/1950
 kvmplutotest       ikev1-policy-03-stats                  good
 
 kvmplutotest	cve-2026-12413-max-payloads		  good
+
+
+#################################################################
+# RFC 9593
+#################################################################
+
+kvmplutotest	ikev2-supported-auth-methods-01                 wip
```
cagney: BTW, good vs wip is up to you.

From my POV, a `good` test can demonstrate an incomplete or even broken feature; just as long as it is stable and commented.  Conversely, `wip` says more about the state of the test then the workings of the feature.

testing/pluto/ikev2-supported-auth-methods-03/west.conf
```diff
@@ -0,0 +1,16 @@
...
+config setup
+	plutodebug="all"
+	logtime=no
+	logappend=no
+	logfile=/tmp/pluto.log
+	dumpdir=/tmp
+
+conn westnet-eastnet-ikev2
+	left=192.1.2.45
+	leftsubnet=192.0.1.0/24
```
cagney, line 12: I suspect you also want to specify rightsubnet=, or just turn this into a host-host config.
As things stand this is trying to negotiate a weird subnet-host config:
  Child SA #2 {ESP <0xESPESP} [192.0.1.0/24===192.1.2.23/32]
which is just a distraction from the juicey part of the test - auth - and explains why the ping fails.

configs/d.ipsec.conf/conn/send-supported-auth-methods.xml
```diff
@@ -0,0 +1,17 @@
...
+  <listitem>
+    <para>
+      Whether to send the SUPPORTED_AUTH_METHODS notification 
+      to the remote peer. When enabled, the locally supported 
+      authentication methods are announced in IKE_SA_INIT response 
+      and IKE_AUTH request, allowing the peer to pick a compatible 
+      method before sending its AUTH payload.
+    </para>
+    <para>
+      The default is &no;.
```
cagney, line 14: Can you just add a note that the feature is experimental.
And cite the relevant RFC.

docbook has some fancy way of adding a caution, but simple text is fine.

We'll just need to remember to replace the text with an historic note about it being added in 5.5 (you could even create a task to remind us :-)

programs/pluto/ikev2_ike_intermediate.c
```diff
@@ -904,6 +911,14 @@ stf_status process_v2_IKE_INTERMEDIATE_request_continue(struct ike_sa *ike,
 		close_pbs_out(&ppks);
 	}
 
+	/* Send the responder's SUPPORTED_AUTH_METHODS notification */
+	if (task->exchange.supported_auth_methods)
+	{
+		if (!emit_v2N_SUPPORTED_AUTH_METHODS(ike, response.pbs)) {
```
cagney, line 916: We need to fix your editor's settings.

Linux puts braces on the same line as the `if`.

configs/d.ipsec.conf/sect/conn.xml
cagney: True.

But honestly, file is a dogs breakfast (we're carefully skirting around this with that doco pull request :-).  The big thing is that it's listed :-)

cagney:
@OsemaFadhel the second half of this pull request comes with no surprises - works for me.
Yes, I nitpicked a little - I'd mainly fix the test.

So once the first round is merged this should rebase and merge pretty quick.

testing/x509/generate.sh
OsemaFadhel: Tested certutil --rename but it doesn't help, because if they have the same subject DN, nss stores them with the same nickname. I'll open a separate pr for this change.

OsemaFadhel: I think a long-term fix would be to update get_local_private_key() to find the cert using both the nickname and the key type.

OsemaFadhel:
> @OsemaFadhel the second half of this pull request comes with no surprises - works for me. Yes, I nitpicked a little - I'd mainly fix the test.
> 
> So once the first round is merged this should rebase and merge pretty quick.

Updated! Also added a new test for IP fragmentation case :)

cagney:
there's a rumour that 5.4 is about to be branched; I'll test but this might be delayed a few days

cagney (approved):
nice work

OsemaFadhel:
> nice work

Thanks :)

PR #2936, opened 2026-07-22
ikev2: omit ADDITIONAL_KEY_EXCHANGE in the last followup-ke exchange

ueno:
The responder should send a link in the ADDITIONAL_KEY_EXCHANGE notification if and only if the next exchange is IKE_FOLLOWUP_KE, according to RFC 9370.

This also includes tests that exercise situations where the notification is omitted from both sides, as well as some minor refactoring.

Fixes: #2926

Issue #2530, opened 2025-11-15
merge refcnt into leak-detective; making leak-detective a requirement


cagney:
### What needs to be done

While leak-detective is always compiled in, enabling it is optional, however all default deployments do enable it.  So time to drop it being optional?

refcnt is currently a layer over the top of the memory allocator.  If leak-detective is made a requirement the two features can be merged.

cagney:
this also allows realloc() to be made more robust

PR #2951, opened 2026-07-30
Fix IKE_FOLLOWUP_KE processing with multiple ADDKE configuration

ueno:
While trying to address review comments in #2936, I noticed that rekeying with multiple ADDKE is not working because of the misuses of table API. This should fix it and expand the test to cover the configuration.


cagney (approved):
nice catch

see #2530, I'd like everything refcnt'd which would hopefully make fixing these problems easier

PR #2968, opened 2026-08-10
Generate certificates for SUPPORTED_AUTH_METHODS tests

OsemaFadhel:
Update generate.sh to generate RSA and ECDSA certificates for east/west (rsa-{east,west}, ecdsa-{east,west}), each with a distinct Subject DN.

Because NSS stores certificates that share the same Subject DN under the same nickname, generating them this way bypasses the limitation.

see #2950


cagney:
Something feels wrong.  I gather `certutil --rename` didn't help.

Not being able to have EC and RSA certs with the same DN strikes me as a strange limitation.  For instance:

- a server migrating clients from RSA to EC will need to build an entirely new DN structure
- a client shouldn't be changing its ID, taken from the DN, just because it has a new cert

Then again, it's NSS.  I just wonder if the cause is elsewhere.  Several things come to mind:

- the RSA and EC certs could have an identical serial number
  there's separate pki/real/{mainec,mainca}/serial when there should be a shared pki/real/serial
- both RSA and EC are generated with the same nicname
  it was done to make switching from RSA to EC easier; but in hindsight ...
  this means that when importing a .p12 there'll be a clash with both the root and end cert so more renaming may be needed

cagney:
I'll try to push a fix for #2971 - the non-unique serial  problem - later today.

cagney:
the plot thickens; the root ec has the nic mainec, and a unique dn, its just the end certs that have name problems

cagney:
rebaseing

cagney (approved):
I'll merge this as is.

I can then investigate the cert problems without holding up your next commit.

PR #2973, opened 2026-08-13
feat: add ike-sa-init-full-transcript-auth= connection parameter

Vinnu124:
Adds the ike-sa-init-full-transcript-auth= keyword and plumbs the value into config.(ikev2 only)
The default is kept as `no`(as of now)


cagney:
@Vinnu124 FYI, I'm going to quickly rebase and force push

cagney (changes requested):
@Vinnu124 two tweaks

- make the default `no` for now; else at some point we'll end up negotiating something that isn't yet working
- add a caution to the documentation (at the top), stating that the feature is experimental and, contrary to the below, it is `no`

then create a bug that we'll need to clean these up

Vinnu124:
@cagney 
Done.
Have updated the docs and the PR description as well

cagney (approved):
yep!
sorry about the confusion over initial defaults.

PR #2978, opened 2026-08-13
building: repair USE_IKEv1=false

paulwouters:
By creating this pull-request you are certifying that your
contribution complies with the [Developer's Certificate of
Origin](https://developercertificate.org/).

_It is very unlikely that a contribution created with the help of AI
complies with this requirement._

When merging, would like to use this pull-request's _Subject_ and
_Description_ (i.e., this text) as part of the Merge commit message!
So please be concise with what you write here!

Please be sure to work through through the following:

- our [Programming
  Conventions](https://github.com/libreswan/libreswan/wiki/Hacking:-Programming-Conventions)
  guideline

- our [Git, GitHub, and Pull
  Requests](https://github.com/libreswan/libreswan/wiki/Hacking:-Git,-GitHub,-and-Pull-Requests)
  guideline

_If you're a GSoC candidate then creating a pull-request is not the
best way to initiate contact with the Libreswan community._

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -77,6 +77,6 @@ jobs:
 
       - name: Perform CodeQL Analysis
         # pinning to full commit hash from tagged version v2.20.4
-        uses: github/codeql-action/analyze@54f647b7e1bb85c95cddabcd46b0c578ec92bc1a
+        uses: github/codeql-action/analyze@5595ccaf912efad79be6eef63a5619ff05969be3
         with:
           category: "/language:${{matrix.language}}"


programs/algparse/algparse.c
```diff
@@ -578,7 +582,9 @@ const struct option optarg_options[] = {
 	{ OPT("verbose"), no_argument, NULL, OPT_VERBOSE, },
 	{ OPT("version"), no_argument, NULL, OPT_VERSION, },
 	{ OPT("debug"), optional_argument, NULL, OPT_DEBUG, },
+#ifdef USE_IKEv1
 	{ OPT("ikev1"), no_argument, NULL, OPT_IKEv1, },
```
cagney, line 586: I think the changes need to be surgical, for instance, tweaking the file with:
```
{ OPT("ikev1"), no_argument, NULL, OPT_IKEv1, },
```
to
```
#ifdef USE_IKEv1
{ OPT("ikev1"), no_argument, NULL, OPT_IKEv1, },
#else
{ FATAL_OPT("ikev1", "6.0"), no_argument, NULL, 0, },
#endif
```
will stop the option.

programs/pluto/connections.c
```diff
@@ -261,7 +264,7 @@ void discard_connection_spds(struct connection *c)
 	}
 	pfreeany(c->child.spds);
```
cagney, line 268: Note this call to `pfreeany()`, along with `sdp_db_del()` above.  This function - `discard_connection_spds()` - can't be deleted.

cagney:
What's the test plan?

My suggestion here is to be more strategic.

- break all the IKEv1 entry points into libreswan such that they log not supported
  + the fewer `#ifdef`s the less likely something is to break
  + the fewer `#ifdef`s the easier it is to maintain dual builds
  + set up the fedora domain to build/test this; in theory `KVM_FEDORA_USE_IKEv1=false`

For instance:
```
$ ipsec addconn --name foo keyexchange=ikev1
"foo": unable to load connection: keyexchange=ikev1 is obsolete
```

- audit the testsuite and confirm that all IKEv1 tests that matter have an IKEv2 equivalent
  + this is why I've been (re)naming tests `blah-blah-blah-{ikev1,ikev2}`
  + *pluto* tests might be IKEv1, might
  + are we expecting IKEv2 tests (well, non IKEv1 tests) to pass with both builds?

This should give us two things:

- an IKEv2 only build we've confidence in
- a framework that lets us disentangle some of the gnarly IKEv1+IKEv2 code

programs/pluto/connections.c
paulwouters: this function is an ikev1 only function getting called in ikev2. might be worth fixing in a separate PR then.

cagney: Um, `discard_connection_spds()` isn't IKEv1 only.

However, It does eventually call:
```
static void discard_spd_end_content(struct spd_end *e)
{
        virtual_ip_delref(&e->virt);
}
```
which is IKEv1 only.  So a separate function, `discard_connection_virtual_ips()` (for want of a better name) handling just that?

paulwouters: yes that is what I meant. The only function of discard_connection_spds() is for ikev1, so none of the ikev2 code should ever call it. Again, that fix should prob be a separate PR but it is in here so that compiling works.

programs/algparse/algparse.c
paulwouters: why does this need to happen in the devel tool argparse ?

include/ietf_constants.h
```diff
@@ -757,7 +757,9 @@ enum ikev1_xauth_type {
 	IKEv1_XAUTH_TYPE_SKEY = 3,
 };
 
-extern const struct enum_names ikev1_xauth_type_names;
+#ifdef USE_IKEv1
+extern const struct enum_names ikev1_xauth_type_names;	/* #ifdef USE_IKEv1 */
+#endif
```
cagney, line 762: My suggestion here is to move the IKEv1 constants - both enums and names declarations - to a separate file, `ietf_ikev1_constants.h` say, that's explicitly included (i.e., not via constants.h).

That way:
- these `#ifdefs` aren't needed
- `#include "ietf_ikev1_constants.h"` is easier to spot than individual fields
- filtering it from `lib/libswan/Makefile:ENUM_INCLUDES` is straight forward

lib/libswan/constants.c
```diff
@@ -873,6 +896,7 @@ enum_names ikev2_cert_type_names = {
 	&ikev2_cert_type_names_2
 };
 
+#ifdef USE_IKEv1
```
cagney, line 899: If the table is strictly IKEv1 then can I suggest moving it to `lib/libswan/names/ikev1_names.c` (or `lib/libswan/names/<name-of-enum>_names.c` more painful).

Make can then exclude the file from the build.

Hopefully that only leaves two enum_enum_names:
```
payload_type_names
exchange_type_names
```
I suspect both of these messes can be eliminated by adding fields to
```
 struct ike_info {
  const struct enum_names *payload_type_names;
  const struct enum_names *exchange_type_names;
};
```
ditto for some other constants

programs/algparse/algparse.c
cagney: It's a documented utility.

programs/pluto/connections.c
cagney: > The only function of discard_connection_spds() is for ikev1

Again, No.  **It is needed by both by IKEv1 and by IKEv2**

programs/algparse/algparse.c
paulwouters: It's a debugging tool? So I dont think it really needs to hold the user's hand.

programs/pluto/connections.c
paulwouters: fixed.

paulwouters:
a test suite audit of ikev1/ikev2 tests / renames feels outside the scope of this PR.

programs/algparse/algparse.c
cagney: anyone asking what a proposal is is pointed at it

cagney:
> a test suite audit of ikev1/ikev2 tests / renames feels outside the scope of this PR.

It's a predicate.  Else we end up with a code base of unknown and unmaintainable quality.

programs/pluto/connections.c
cagney: Presumably, .virt is IKEv1 only; and doesn't exist when no-USE_IKEv1?  Hence I'd reduce it to:
```
static void discard_spd_content(struct spd *spd)
{
	spd_db_del(spd);
#ifdef USE_IKEv1
	FOR_EACH_THING(end, LEFT_END, RIGHT_END) {
		virtual_ip_delref(&e->virt)(&spd->end[end].virt);
	}
#endif
}
```
There's also no reason to sit on this?

PR #3003, opened 2026-08-21
IKEv2: crossed streams IKE SA should be deleted without family

letoams:
While this does not affect the testcase, it seems the more right thing to do?


cagney:
An IKEv2 Child SA can't out live its parent?

cagney:
Yea, this isn't right.  IKEv2 Child SAs can't out live their IKE SA

PR #3017, opened 2026-08-31
ikev2: fall back to local auth config when negotiated_authby is empty (RFC 9593)

OsemaFadhel:
When processing a peer's SUPPORTED_AUTH_METHODS notification (RFC 9593), if the peer announces methods that do not match our local configuration (e.g., the peer announces ECDSA but we only have PSK), the negotiation results in 0 (none).

This triggered an assertion failure in local_v2AUTH_method():
```
Aug 27 21:15:42.102991: | SUPPORTED_AUTH_METHODS: selecting negotiated authby=0
Aug 27 21:15:42.102996: FATAL: ASSERTION FAILED: switch (authby) case 0 (0x0) unexpected (local_v2AUTH_method() +299 programs/pluto/ikev2_auth.c)
```

This commit adds a fallback so that if no shared auth method is found, it safely falls back to using the locally configured auth method:

```
Aug 27 22:21:53.633076: | SUPPORTED_AUTH_METHODS: no negotiated authby, falling back to local config: secret
```


cagney:
is there a way to test this?

the, if all else fails strategy is add an impair; but there may be another way

cagney:
btw, impair knows about these enums
```
enum impair_emit {
        IMPAIR_EMIT_NO = 0,
        IMPAIR_EMIT_OMIT,
        IMPAIR_EMIT_EMPTY,
        IMPAIR_EMIT_DUPLICATE,
#define IMPAIR_EMIT_ROOF (IMPAIR_EMIT_DUPLICATE+1) /* >= ROOF -> <number> */
};
```
there's no expectation to use all of them, but might be useful

cagney (approved):
nice

PR #3036, opened 2026-09-12
feat: carry full-transcript-auth through session resume

Vinnu124:
Save the full-transcript-auth decision in the resume ticket and restore it on both ends when resuming, since IKE_SESSION_RESUME never renegotiates it.


cagney (approved):
nothing scary here!

